PR #23, opened 2017-05-02
Code cleanup, implement Notifications example db hookup

GeorgeWalker:


matthewhall78:
[(5) Back End Development: Extend current Django implementation to use the Django REST Framework [5]](https://trello.com/c/mYKwO0aV/58-5-back-end-development-extend-current-django-implementation-to-use-the-django-rest-framework-5)

PR #24, opened 2017-05-11
Jenkins Pipeline (init)

dmarley:
You will need to update the GIT URL on line four of Jenkinsfile to the TFRS git url


Jenkinsfile
```diff
@@ -0,0 +1,26 @@
+node('maven') {
+
+   stage('checkout tfrs src onto docker slave')
+   git url: 'https://github.com/dmarley/tfrs.git'
```
cvarjao: shouldn't be using `checkout scm`? it looks like it will checkout from a fork as opposed to the officila/bcgov repository

PR #32, opened 2017-06-29
Enabling generation of Django code from swagger yaml file and check i…

swcurran:
…n of the generated code.

This is the task on the Trello board to put the generated code into the repo for comparison on future generation runs.


matthewhall78:
[(0.5) Update the codegen process/documentation to include storing, reviewing the generated-not-used files to detect changes](https://trello.com/c/wWEg9WHu/79-05-update-the-codegen-process-documentation-to-include-storing-reviewing-the-generated-not-used-files-to-detect-changes)

PR #48, opened 2017-07-25
add audit functionality

jeffmcnd:
This commit adds a folder which contains the files responsible for performing the auditing on models in the database using the Note model as an example.


matthewhall78:
Will you be submitting a replacement pull request?

PR #131, opened 2017-11-29
Ticket 332. For the purposes of creating realistic mockups, I've move…

rajpersram:
For ticket 332 - Create working, portable mockups: https://trello.com/c/koD9dkdH. Creating initial commit of static files.


rajpersram:
Just to clarify, the original files were just the layout and don't resemble the up to date semi-functional working mockup. That's why I moved them into a v1 directory, for archiving. The v1.1 represents interactive functionality based on how our app is today and as a result the structure (filenames, organization) is very different so versioning between v1 and v1.1 does not make sense (apples & oranges). However, moving forward, versioning from 1.1 to future working mockups does make sense (different types of oranges). Perhaps a better naming of the directories would be: 'archive' and 'interactive' instead of versions.

rajpersram:
Whatever works for you guys. For the record, I wasn't looking to version the initial ones at all because the new files are not related to them. 'Demo' works as a naming convention for me.

dainetrinidad:
Please include the deletes in your commit as well (right now there are no deletes in this PR).

dainetrinidad:
@rajpersram just to confirm, can you make the following changes:

* Rename the `Wireframe` folder to `wireframes`
  * @brian-22 Once this PR is approved & merged, can you update the links in the wiki that point to this folder?
* Delete all other files and folders.
* Create a `demo` folder where you will place your current html mockups into.

rajpersram:
Will do.

PR #135, opened 2017-12-14
Upgrade to BDDStack

rstens:
Just the Test Framework so far. The update to the JenkinsFile still needs to be made.
Also the BDDStack Node must be defined in Jenkins.


matthewhall78:
[Testing | create a BDD framework such that additional tests can be created/executed) |0|](https://trello.com/c/shNEkhu5/62-testing-create-a-bdd-framework-such-that-additional-tests-can-be-created-executed-0)

PR #140, opened 2017-12-15
Authentication

dainetrinidad:
This PR adds/enables authentication to the API.

- Use siteminder headers to authenticate user
- Use Django's authentication module
  - Extend Django's AbstractUser class
  - Changed two migration files because deciding to use Django's auth module after migrations have been created is tricky (see https://docs.djangoproject.com/en/1.11/topics/auth/customizing/#changing-to-a-custom-user-model-mid-project)
- Create user from headers sent by siteminder if their user guid doesn't exist in the database yet
- All tests would fail once the authentication is set in place as it was previously set to "AllowAny", so those are changed in this PR as well.
- Enabled Django Admin UI for super users (developers, maintainers etc)


kuanfandevops:
Kuan's peer reviewed passed.

PR #153, opened 2018-01-19
 Updates to the Jenkinsfiles, these updates will show the test results...

rstens:
in Jenkins. This is easier as you do not require to navigate into jenkins to see them.
We're still archiving all the results as before.

### DIFF
--- a/frontend/Jenkinsfile
+++ b/frontend/Jenkinsfile
@@ -36,20 +36,39 @@ node('maven') {
 }
 
 node('bddstack') {
-  stage('FT on Dev') {
-    echo "checking out source"
-    echo "Build: ${BUILD_ID}"
-    checkout scm
-    dir('functional-tests') {
-      try {
-          sh './gradlew --debug --stacktrace chromeHeadlessTest'
-      } finally {
-          archiveArtifacts allowEmptyArchive: true, artifacts: 'build/reports/**/*'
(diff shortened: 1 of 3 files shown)


frontend/Jenkinsfile
```diff
@@ -36,20 +36,39 @@ node('maven') {
...
-    dir('functional-tests') {
-      try {
-          sh './gradlew --debug --stacktrace chromeHeadlessTest'
-      } finally {
-          archiveArtifacts allowEmptyArchive: true, artifacts: 'build/reports/**/*'
-          archiveArtifacts allowEmptyArchive: true, artifacts: 'build/test-results/**/*'
-          junit 'build/test-results/**/*.xml'
-      }
+
+	stage('FT on Dev') {
```
dainetrinidad: No issues with the changes, but is it possible to rename this stage to something more meaningful? "FT on Dev" doesn't immediately convey "Functional Tests" to me at first glance. It does, now, but for new people it can be confusing since this is the one that shows up on the Openshift web UI as well. Thoughts @kuanfandevops ?

kuanfandevops: Openshift pipeline has limited spaces for each stage, I'd like to rename it as Functional Test, just remove the environement name as "deploy on dev" is the stage right before Functional Test. what do you think?

PR #160, opened 2018-02-06
Updated Openshift templates

kuanfandevops:
Branch can be removed after merging.


openshift/templates/dev/dev-secret.json
```diff
@@ -0,0 +1,20 @@
...
+    "metadata": {},
+    "items": [
+        {
+            "kind": "Secret",
+            "apiVersion": "v1",
+            "metadata": {
+                "name": "tfrs-postgresql",
+                "creationTimestamp": null
+            },
+            "data": {
```
dainetrinidad: Can you connect with the other devops people if this encryption is good enough for storing in the repo?

kuanfandevops:
yes, Secrets shouldn't be in repo even they are encrypted. I have removed them and will update them in dagtabase as well. Thanks for the reminding.

PR #182, opened 2018-02-23
Install ZAP scripts for security scanning

rstens:
Adding the scripts and session information for scanning TFRS. This script will automatically login through siteminder (id/pw are encrypted, so no worries) and run a scan on the TFRS application.

I've added the jenkinsfile that contains the scanning step.


kuanfandevops:
You are awesome Roland! Only one comment from me, does zap/TFRS.session.data need to be in repo? It seems a binary file contains ZAP results or I might be wrong...

kuanfandevops:
The zap/TFRS.session.data is needed, it is an actual hsql database file that is required. 
@dainetrinidad the changes is on frontend but nothing related frontend code. They are all Zap test related. I'll merge the PR.

Issue #206, opened 2018-04-04
Historical Data | Load New Fuel Suppliers |2|


amichard:
**Additional Companies to add:**
Justin provided Daine with an initial list of companies to load into the system. Since that time the Ministry has become aware of several new suppliers that will need to be added to the system. 

As of March 28, 2018, the following companies need to be added (this is in addition to the original list Justin gave Daine): 
1. Teck Coal Limited
2. Parkland Fuel Corporation
3. Elbow River Marketing Ltd. 

https://trello.com/c/mudO188m/554-historical-data-load-new-fuel-suppliers-2

amichard:
This is done.

PR #207, opened 2018-04-04
Added new fuel suppliers

amichard:
#206 

As mentioned in the Trello card 554, we needed to add the following suppliers.

So this merge request is supposed to accommodate that.


kuanfandevops:
Peer reviewed. Approve the change.

PR #196, opened 2018-03-21
Historical data frontend

amichard:
This is the front-end for https://trello.com/c/QVNn229c/533-historical-data-creating-form-to-enter-historical-transactions-8

Just some notes:
- I tried to follow the mock-up as best as I can
- It's currently loading all credit transfers, instead of just approved.
- There are no functionalities that tie up to the backend (commit and add to queue)
- Once I added the backend routes and made some modifications, I should be able to tie them up.

PR #210, opened 2018-04-10
Historical data frontend

amichard:
#196 

Follow-up pull request. 

Changelog:
# Frontend
- Added validation logic for Zero Reason buttons
  - If a zero reason was selected, then conditions were met that should disable it. Button should no longer show a disabled "pressed" button
  - Zero reason button will now be disabled, unless type is Credit Transfer (or no type is selected) and dollar per credit is not greater than 0
- Change sorting for drop-down so it's alphabetical
- Created a helper function to format the data before sending it to the server to reduce duplicated code
- Fixed the issue where it sometimes show two drop downs for Credit From or Credit To
- In the table, prices are only shown for credit transfers
- Credit From and Credit To show N/A in the table where it's appropriate
# Backend
- Should now automatically create a credit balance of 0 for fuel suppliers that didn't have a record prior. This is so that it's guaranteed that the fuel supplier has a credit balance and won't throw an error that the user can't do anything about
- When committing, the process should now validate and make sure that there's sufficient credits for everyone, before committing (it does a check on the current balance so there's still a chance of doing partial commits if it's a transaction in the middle that will cause the fuel supplier to not have enough credits)


backend/api/services/CreditTradeService.py
```diff
@@ -139,21 +139,23 @@ def approve(credit_trade):
     @staticmethod
     def transfer_credits(_from, _to, credit_trade_id, num_of_credits,
```
Sybrand: Isn't it easy to just wrap this in a database transaction?

PR #211, opened 2018-04-10
Upgrade Gunicorn and include config file

kuanfandevops:
TFRS prod keeps having issues which gunicorn starts too many worker threads. The added configuration limits the number threads to be 10.

### DIFF
--- a/backend/gunicorn.cfg
+++ b/backend/gunicorn.cfg
@@ -1,0 +1,50 @@
+# Gunicorn configuration file.
+
+# Worker processes
+#
+#   workers - The number of worker processes that this server
+#       should keep alive for handling requests.
+#
+#       A positive integer generally in the 2-4 x $(NUM_CORES)
+#       range. You'll want to vary this a bit to find the best
+#       for your particular application's work load.
+#
+#   worker_class - The type of workers to use. The default
+#       sync class should handle most 'normal' types of work
+#       loads. You'll want to read
+#       http://docs.gunicorn.org/en/latest/design.html#choosing-a-worker-type
+#       for information on when you might want to choose one
+#       of the other worker classes.
+#
+#       A string referring to a Python path to a subclass of
+#       gunicorn.workers.base.Worker. The default provided values
+#       can be seen at
+#       http://docs.gunicorn.org/en/latest/settings.html#worker-class
+#
+#   worker_connections - For the eventlet and gevent worker classes
+#       this limits the maximum number of simultaneous clients that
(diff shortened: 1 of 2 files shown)


backend/gunicorn.cfg
```diff
@@ -0,0 +1,50 @@
...
+#
+#   keepalive - The number of seconds to wait for the next request
+#       on a Keep-Alive HTTP connection.
+#
+#       A positive integer. Generally set in the 1-5 seconds range.
+#
+
+workers = 10
+worker_class = 'sync'
+worker_connections = 1000
```
Sybrand: This equates to 10000 simultaneous connections. If that isn't enough for TFRS there's definitely some other issue.

backend/gunicorn.cfg
```diff
@@ -0,0 +1,50 @@
...
+#       process is still communicating and is not tied to the length
+#       of time required to handle a single request.
+#
+#   keepalive - The number of seconds to wait for the next request
+#       on a Keep-Alive HTTP connection.
+#
+#       A positive integer. Generally set in the 1-5 seconds range.
+#
+
+workers = 10
```
Sybrand: So I'm assuming TFRS is running on a machine with 2 to 5 cores?

backend/gunicorn.cfg
kuanfandevops: I prefer to use async workers gevent instead of sync worker now. what do you think?

Sybrand: I think that with the load we should be dealing with, it shouldn't actually matter that much. Our only IO bound processing is on the database, which should be pretty zippy. Sync or async should both work find, and neither should be causing any issues. Even if our DB stuff is slow, I doubt we're writing clean non-blocking code to take full advantage of async.

It might be a slightly academic discussion at this point, and not worth the effort, but  one could hit some of the api endpoints with jmeter and compare sync vs. async for some real numbers.

That said - I think we should be striving for async.

kuanfandevops: Agreed. The only "external" access is database which isn't really classified as external. So far TFRS doesn't have any external endpoints to access, so I'll leave worker_calss as sync. The worker_connections default is 1000 and only affects Eventlet and Gevent, I'll remove it. 

In gunicorn doc, it also recommends  2 to 4 * number-of-cores for worker, I'll update it to 8 as I'd like to give 2 cores for prod.

Thanks for all the recommendations!

kuanfandevops:
changes have been reviewed and discussed.

PR #212, opened 2018-04-11
Removed test code

amichard:
Follow-up Pull Request.

Just removing the print which the testing is not liking.


kuanfandevops:
Peer reviewed.

Issue #216, opened 2018-04-12
Bug | Refresh Error


amichard:
https://trello.com/c/csTftYAW/565-bug-refresh-error

Notice that when you hit refresh in /admin/historical_data_entry it causes the site to just show a blank page. The same can be applied if you view a specific credit transaction and refresh.

PR #217, opened 2018-04-12
Changed to use absolute files for asset files

amichard:
#216 

The issue might not be related to the server after all. 

Just changing the path to asset files so it uses absolute instead of relative. This will hopefully fix the blank screen error.


kuanfandevops:
Peer reviewed passed, replace with absolute path.

Issue #225, opened 2018-04-18
Logout | Implement the ability for a user to log out of TFRS |2|


amichard:
https://trello.com/c/FhXyLzlX/418-logout-implement-the-ability-for-a-user-to-log-out-of-tfrs-2

Story:
As a TFRS user, I want to be able to log out of the system so that I can properly and securely end my TFRS session when I am done. 

Scenario:
GIVEN that I am a TFRS user
WHEN I am logged into the system
THEN I can see a log out button in the TFRS Toolbar at the top of every page
AND clicking the log out button securely logs me out of TFRS

Demo: 
Show the log out button in the Toolbar
Show a user logging out

Issue #224, opened 2018-04-18
https://trello.com/c/qn9VS3u6/514-ux-refactor-hide-some-tfrs-tabs-and-contents-2-permissions


amichard:
https://trello.com/c/qn9VS3u6/514-ux-refactor-hide-some-tfrs-tabs-and-contents-2-permissions

**Story:**
As a TFRS PO, I want certain tabs / pages to be hidden to BCeID org type users so that fuel suppliers cannot complete actions they are not supposed to in the initial release (0.2.0). 

**Demo:**
Show that the different tabs are not visible to a BCeID user

Issue #223, opened 2018-04-18
UX Refactor | Add Organization Credit Balance on the Credit Transactions page |3|


amichard:
https://trello.com/c/aFucGmwM/556-ux-refactor-add-organization-credit-balance-on-the-credit-transactions-page-3

**Story:**
As a Fuel Supplier user, I want to be able to quickly and easily see my organization's credit balance on the Credit Transactions page so that I can understand my company's compliance situation. 

**Scenario:**
GIVEN that I am a Fuel Supplier user 
WHEN I am logged into TFRS
THEN I can see my organization's credit balance in the right-hand side of the Credit Transactions page

**Demo:**
Show that an organization's credit balance is visible on the right hand side of the Credit Transactions page

Issue #222, opened 2018-04-18
UX Refactor | Hide New Transfer Button on Credit Transactions Page |1|


amichard:
https://trello.com/c/DtBncpEB/555-ux-refactor-hide-new-transfer-button-on-credit-transactions-page-1

**Story:**
As a TFRS PO, I want to hide the 'New Transfer' Button on the Credit Transactions Page so that when we have our first release, components of TFRS that are still in-progress are not visible to external users. 

**Scenario:**
GIVEN that I am a Fuel Supplier user 
WHEN I navigate to the Credit Transactions page
THEN the 'New Transfer' button is NOT visible on the top left-hand side of the page (where the Propose Trade link is right now)

**Demo:**
Show that the 'New Transfer' button is not visible to external users (i.e. fuel suppliers) on the Credit Transactions page

PR #226, opened 2018-04-18
UX Refactor

amichard:
#221 
#222 
#223 
#224 
#225 

This pull request focuses mostly on the front-end. Below is a summary of the changes:
- Replaced Propose Trade link with button
- Added Credit Balance in the Credit Transactions page
- Added conditions on when to display the Propose Trade Button and tabs other than Credit Transactions
- Credit Transaction view details updates
  - Improvements on how price and number of credits is displayed
  - Improvements on the text being displayed so its more relevant to the type of transfer
- Minor refactoring with the routes
- Code clean-up for Lint


backend/api/serializers.py
```diff
@@ -190,13 +190,26 @@ class Meta:
...
     class Meta:
         model = User
         fields = (
             'id', 'first_name', 'last_name', 'email', 'authorization_id',
             'authorization_guid', 'authorization_directory', 'display_name',
-            'organization')
+            'organization', 'organization_balance')
+
+    def get_organization_balance(self, obj):
+        try:
```
Sybrand: Maybe a bit pedantic, but if this is not an exceptional circumstance, one shouldn't be handling it as an exception, but rather as a check. I would expect exception handling to be involved if something is actually wrong, and it be logged etc.

frontend/src/credit_transfers/components/CreditTransferTextRepresentation.js
```diff
@@ -0,0 +1,145 @@
...
+    this.totalValue = numeral(this.props.totalValue).format(NumberFormat.CURRENCY);
+    this.tradeEffectiveDate = (this.props.tradeEffectiveDate)
+      ? moment(this.props.tradeEffectiveDate).format('LL') : 'approval';
+
+    this.tradeStatus = (this.props.status.id === CREDIT_TRANSFER_STATUS.completed.id)
+      ? CREDIT_TRANSFER_STATUS.approved.description : this.props.status.status;
+  }
+
+  _buyAction () {
+    if (this.props.status.id === CREDIT_TRANSFER_STATUS.completed.id) {
```
Sybrand: Just a personal preference note: I would do:
`return this.props.status.id === CREDIT_TRANSFER_STATUS.completed.id ? ' bought ' : ' is proposing to buy '`

I think it's easier to read, and your function only has one exit point.

frontend/styles/CreditTransfers.scss
```diff
@@ -69,20 +64,22 @@
     
       .step {
         font-size: 14px;
+        font-weight: bold;
         text-align: center;
-        color: $light-font;
+        color: $medium-font;
```
Sybrand: 😀 woohoo SASS! Loving it. Try to guard against nesting too deeply though.

frontend/src/credit_transfers/components/CreditTransferTextRepresentation.js
amichard: sounds good. i have no problems with ternary functions.

Issue #243, opened 2018-04-26
Security | Measures to stop externals from sending requests in prod [v0.2.0]


amichard:
https://trello.com/c/Q9VjElhJ/588-security-measures-to-stop-externals-from-sending-requests-in-prod-v020

Implement measures to stop externals (BCeID org types) from sending requests in prod (e.g. credit transfer requests). Although most features are hidden to BCeID org types for the initial releases (v0.1.0 & v0.2.0), there are not measures in place to stop externals from sending requests.

PR #244, opened 2018-04-26
Authorization approve

amichard:
#243

This is to ensure that the users can't approve their own credit transfers. So I'm adding a check that the user who's approving is associated with the government.


backend/api/services/CreditTradeService.py
```diff
@@ -112,7 +115,13 @@ def create_history(credit_trade, is_new=False):
         history.save()
 
     @staticmethod
-    def approve(credit_trade):
+    def approve(credit_trade, user):
+        """Only government users can approve transfers"""
+        gov_org_type = OrganizationType.objects.get(type="Government")
+
+        if user.organization.type_id != gov_org_type.id:
```
Sybrand: I think this should be happening on the view/api level, not in this helper function. The API should be responsible for enforcing the contract. I think it should also rather make use of the built in django permissions, then it can be enforced by adding a decorator on the endpoint.

amichard:
[Security | Measures to stop externals from sending requests in prod [v0.2.0]](https://trello.com/c/Q9VjElhJ/588-security-measures-to-stop-externals-from-sending-requests-in-prod-v020)

Issue #259, opened 2018-05-01
Confirmation Alerts | Delete a Saved Draft of a Credit Transfer |1|


amichard:
https://trello.com/c/LP1JDWhu/500-confirmation-alerts-delete-a-saved-draft-of-a-credit-transfer-1

**Story 1: **
As a Fuel Supplier Initiator, I want the system to confirm a deletion when I try to delete a Saved Draft so that I don't accidentally delete a draft credit transfer. 

**Scenario 1:**
GIVEN that I am a Fuel Supplier user
AND I am viewing a saved draft of a Credit Transfer Proposal
WHEN I click on Delete
THEN I receive a confirmation alert that asks me to confirm that I want to Delete the Draft
AND I can either confirm deletion or cancel/go back

**Demo:**
Show the confirmation alert when deleting a saved draft of a Credit Transfer Proposal
Show that the user can either confirm or cancel/go back

Issue #258, opened 2018-05-01
Credit Transfer | Initiator can Delete a credit transfer |3|


amichard:
https://trello.com/c/PZKSM4Rd/324-credit-transfer-initiator-can-delete-a-credit-transfer-3

**Story:**
 As a Fuel Supplier Initiator, I want to be able to delete a draft of a credit transfer proposal so that my colleague doesn't mistakenly propose the transfer to the other Fuel Supplier. 

**Scenario:**
GIVEN that I am an Organizational user
AND my Organization type is Fuel Supplier
AND I have saved a draft of an offer to purchase 5,000 credits at $160 per credit from Electric Drive Canada
AND I am viewing that saved draft
WHEN I click Delete
THEN I am returned to the Credit Transactions page
AND the transaction history record is removed from the Credit Transactions page

**Demo:**
Show that Draft transfers can be deleted by Fuel Supplier Initiators
Show that Credit Transfer Proposals that do not have a Draft status can't be deleted

PR #260, opened 2018-05-01
Added Delete function in Edit Form.

amichard:
#258 
#259 

We already had a delete function before. So it's a matter of putting the confirmation and having it show up in the edit page

Changelog:
- Delete button in Edit Draft
- Minor refactoring
- Changed the modal message and the buttons
- Credit transfers page should no longer show "Cancelled" credit transfers

### DIFF
--- a/backend/api/test_credit_trades.py
+++ b/backend/api/test_credit_trades.py
@@ -81,7 +81,7 @@ class TestCreditTrades(TestCase):
     # I'm the respondent, if the status is "submitted" or greater
     def test_initiator_should_see_appropriate_credit_trades(self):
         response = self.fs_client_1.get('/api/credit_trades')
-        assert response.status_code == status.HTTP_200_OK
+        self.assertEqual(response.status_code, status.HTTP_200_OK)
 
         fs_credit_trades = response.json()
         for ct in fs_credit_trades:
@@ -91,7 +91,7 @@ class TestCreditTrades(TestCase):
             elif (ct['respondent']['id'] == self.user_1.organization.id and
                   ct['status']['id'] >= STATUS_SUBMITTED):
                 correct_view = True
-            assert correct_view is True
+            self.assertTrue(correct_view)
 
     # As a government user, I should see all credit trades where:
     # I'm the initiator, regardless of status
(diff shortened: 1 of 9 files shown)


backend/api/test_credit_trades.py
```diff
@@ -439,3 +439,46 @@ def test_batch_process(self):
...
+            type=credit_trade_type,
+            number_of_credits=1000,
+            fair_market_value_per_credit=0,
+            zero_reason=credit_trade_zero_reason,
+            trade_effective_date=datetime.datetime.
+            today().strftime('%Y-%m-%d'))
+
+        response = self.gov_client.put('/api/credit_trades/{}/delete'.format(
+            credit_trade.id))
+        assert response.status_code == status.HTTP_200_OK
```
Sybrand: Please use self.assertEqual , not assert for unit tests. self.assertEqual will give better error message. Also, more important, when not running in debug mode, these asserts won't work correctly, as relies on __debug__ being True. (See https://docs.python.org/3/reference/simple_stmts.html#assert )

backend/api/test_credit_trades.py
```diff
@@ -439,3 +439,46 @@ def test_batch_process(self):
...
+
+        credit_trade = CreditTrade.objects.get(id=credit_trade.id)
+
+        assert credit_trade.status_id == cancelled_status.id
```
Sybrand: Have a look at the "arrange, act, assert" design pattern.

amichard:
[Confirmation Alerts | Delete a Saved Draft of a Credit Transfer |1|](https://trello.com/c/LP1JDWhu/500-confirmation-alerts-delete-a-saved-draft-of-a-credit-transfer-1)

amichard:
[Credit Transfer | Initiator can Delete a credit transfer |3|](https://trello.com/c/PZKSM4Rd/324-credit-transfer-initiator-can-delete-a-credit-transfer-3)

backend/api/test_credit_trades.py
amichard: Sure. I can change asserts into self.assertEqual instead

Issue #246, opened 2018-04-27
Filtering | Fix issues with the search / filter function in the Credit Transactions Table |5| [v0.2.0]


amichard:
https://trello.com/c/GFgQcEOu/568-filtering-fix-issues-with-the-search-filter-function-in-the-credit-transactions-table-5-v020

There appears to be several issues with the filtering function in the Credit Transactions Table that will need to be resolved prior to the v0.2.0 release

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Filtering | Fix issues with the search / filter function in the Credit Transactions Table |5| [v0.2.0]](https://trello.com/c/GFgQcEOu/568-filtering-fix-issues-with-the-search-filter-function-in-the-credit-transactions-table-5-v020)

PR #264, opened 2018-05-02
Follow-up fix for the currency filter. This allows decimal filtering

amichard:
#246 

I noticed that when I tried to add .00 on the currency column, it removes the matching rows. This was caused by the values in the accessor had no decimal places when it ends with .00

### DIFF
--- a/frontend/src/admin/historical_data_entry/components/HistoricalDataTable.js
+++ b/frontend/src/admin/historical_data_entry/components/HistoricalDataTable.js
@@ -76,7 +76,7 @@ const HistoricalDataTable = (props) => {
         return -1; // this is to fix sorting (value can't be negative)
       }
 
-      return parseFloat(item.totalValue);
+      return parseFloat(item.totalValue).toFixed(2);
     },
     className: 'col-price',
     Cell: row => (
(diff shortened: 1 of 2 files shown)


frontend/src/admin/historical_data_entry/components/HistoricalDataTable.js
```diff
@@ -76,7 +76,7 @@ const HistoricalDataTable = (props) => {
         return -1; // this is to fix sorting (value can't be negative)
       }
 
-      return parseFloat(item.totalValue);
+      return parseFloat(item.totalValue).toFixed(2);
```
Sybrand: There aren't any business rules/legislation on how we round is there?

amichard: what do you mean? I used 2 decimal places for the reason that that's what's closes to what the user will see in the table.

Sybrand: Financial systems often have rules around rounding up and rounding down - I'm guessing it's not applicable but given that there's a financial impact with any numbers we're dealing with it may be applicable.

amichard: I'll have to confirm this with Justin

PR #270, opened 2018-05-03
This fixes the sorting and the filter for credit balance

amichard:
#246 

Changelog:
- Moved the formatting outside of accessor so the user can still filter without needing to put in comma. It also fixes sorting


amichard:
[Filtering | Fix issues with the search / filter function in the Credit Transactions Table |5| [v0.2.0]](https://trello.com/c/GFgQcEOu/568-filtering-fix-issues-with-the-search-filter-function-in-the-credit-transactions-table-5-v020)

Issue #272, opened 2018-05-07
Allow users to enter .00 in Tx window |1| [v0.2.0]


amichard:
https://trello.com/b/O9L8CXes/transportation-fuels-reporting-system

**Story:**
As a Fuel Supplier credit trader, I want to be able to record cents (.00) when I am creating or editing credit transfers because some of our agreements are lump sum amounts that do not divide into whole dollar amounts when calculating the value per credit. 

**Scenario:**
GIVEN that I am a Fuel Supplier
AND my role is Credit Trader
AND I am either creating a new transfer or editing an existing saved draft transfer
WHEN I enter the value per credit for the transfer
THEN I am able to specify the exact value down to the cent (e.g. $154.88)

**Demo:**
Show that a Fuel Supplier user (i.e. Credit Trader) can specify the value per credit down to the cent (e.g. $154.88) when either creating a new transfer or editing an existing saved draft transfer

PR #273, opened 2018-05-07
Allow Decimals in Credit Transfer Add/Edit Forms

amichard:
#272 

This should allow decimals for the credit value.

Changelog:
- Removed fixed .00 read-only field
- Replaced parseInt with parseFloat to allow decimals


amichard:
[Allow users to enter .00 in Tx window |1| [v0.2.0]](https://trello.com/c/L8aW8xAt/484-allow-users-to-enter-00-in-tx-window-1)

Issue #286, opened 2018-05-11
Credit Transactions Table | Hide transactions that have an Approved status in the back-end |1| [v0.2.0]


amichard:
https://trello.com/c/4kgdeGFy/595-credit-transactions-table-hide-transactions-that-have-an-approved-status-in-the-back-end-1-v020

As a user, This card is to hide Credit Transactions with an Approved status (Approved in the back-end). Right now, when a credit transaction is added to the queue in the historical data entry, it appears in the Credit Transactions table as "Approved". We want to hide these entries until they have actually been committed in case there are edits made to the entries whilst in the queue. Once the credit transaction is committed, the status will change to "Completed" in the back-end (Approved in the front-end UI), and will show up in the Credit Transactions table

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Credit Transactions Table | Hide transactions that have an Approved status in the back-end |1| [v0.2.0]](https://trello.com/c/4kgdeGFy/595-credit-transactions-table-hide-transactions-that-have-an-approved-status-in-the-back-end-1-v020)

PR #287, opened 2018-05-11
Hide Credit Transactions that are "Approved"

amichard:
#286 

The Credit transactions table should show the relevant records as a government user or as a fuel supplier. It shouldn't show "Approved" ones as it's a status in the historical data entry page indicating that it's ready to be committed, but not "Completed" yet. (Though, front-end will show "Approved" as "Completed" anyway). In other words, "Approved" transactions don't have value for Fuel Suppliers until they're actually committed (As "Approved" hasn't transferred credits yet)

Changelog:
- Changed how the filter works for get_organization_credit_trades
- More unit tests


amichard:
[Credit Transactions Table | Hide transactions that have an Approved status in the back-end |1| [v0.2.0]](https://trello.com/c/4kgdeGFy/595-credit-transactions-table-hide-transactions-that-have-an-approved-status-in-the-back-end-1-v020)

backend/api/test_credit_trades.py
```diff
@@ -483,3 +483,216 @@ def test_delete(self):
...
+            fair_market_value_per_credit=1,
+            zero_reason=None,
+            trade_effective_date=datetime.datetime.
+            today().strftime('%Y-%m-%d'))
+
+        credit_trades = CreditTradeService.get_organization_credit_trades(
+            self.gov_user.organization
+        )
+
+        self.assertNotIn(draft_credit_trade, credit_trades)
```
Sybrand: I would have broken each one of these into a separate test. I would create the CreditTradeStatus and CreditTradeType needed as part of the setup, and then have a separate test for each of the assertions.

backend/api/test_credit_trades.py
```diff
@@ -483,3 +483,216 @@ def test_delete(self):
...
+        credit_trades = CreditTradeService.get_organization_credit_trades(
+            test_organization_1
+        )
+
+        self.assertNotIn(draft_credit_trade, credit_trades)
+        self.assertIn(draft_credit_trade_from_fuel_supplier, credit_trades)
+        self.assertNotIn(submitted_credit_trade, credit_trades)
+        self.assertIn(submitted_credit_trade_as_respondent, credit_trades)
+        self.assertNotIn(approved_credit_trade, credit_trades)
+        self.assertIn(completed_credit_trade, credit_trades)
```
Sybrand: Same comment as above. I don't think there's a good reason to bunch so many assertions in a single test. I think one should default to the "one assertion per test" approach unless there's a really good reason.

backend/api/test_credit_trades.py
amichard: The problem with breaking these into separate tests is that it increases the chance of something going wrong. 

The aim of the test is to check and make sure that certain statuses show up and certain statuses to not show up when calling a specific function.

Sybrand: Sure, there's always a trade-off like that when splitting up closely related tests, if you feel the trade-off isn't worth it, then it probably isn't.

Issue #288, opened 2018-05-15
Signing Authority | Manage versions of assertions |3|


amichard:
https://trello.com/c/NHbWNn0V/609-signing-authority-manage-versions-of-assertions-3

- Create a Table for the Assertions
- Replace the hardcoded assertions with the ones from the table

PR #289, opened 2018-05-15
Signing authority models

amichard:
#288 

I apologize in advanced. I decided to do some refactoring so hence the massive file changes. The serializer being in one place was starting to become unruly so I decided to split them by models.

Changelog:
- Added SigningAuthorityAssertion Table
- Added SigningAuthorityConfirmation Table
- Created model for SigningAuthorityAssertion
- Replaced the hard-coded assertions with the ones from the table
- Checkbox should now call a function to automatically add the field into the state
- Refactored Serializer


amichard:
[Signing Authority | Manage versions of assertions |3|](https://trello.com/c/NHbWNn0V/609-signing-authority-manage-versions-of-assertions-3)

PR #300, opened 2018-05-23
Restored additional parenthesis

amichard:
#299 

I didn't know that the additional parenthesis was that important. So I'm restoring it.


amichard:
[Confirmation Alerts | Success Message doesn't disappear |1| [v0.3.0]](https://trello.com/c/OddDUhYE/620-confirmation-alerts-success-message-doesnt-disappear-1-v030)

Issue #301, opened 2018-05-23
Confirmation Alerts | Success Message doesn't disappear


amichard:
https://trello.com/c/OddDUhYE/620-confirmation-alerts-success-message-doesnt-disappear

When you navigate away from the Administration page and then return to it, the Confirmation Alerts should be cleared or reset back to their defaults

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Confirmation Alerts | Success Message doesn't disappear |1| [v0.3.0]](https://trello.com/c/OddDUhYE/620-confirmation-alerts-success-message-doesnt-disappear-1-v030)

PR #302, opened 2018-05-23
Success/Error Messages clear

amichard:
#301 

There's some instances where the error or success message don't go away. Such as after processing the approved transfers. Another would be if you get a validation error message in the propose trade screen.

Changelog:
- Refactored on where the Action Types constants are stored.
- Success Message from Historical Data Entry should clear properly
- Error Message in Propose trade should clear properly
- Fixed an issue where View Credit Transfer would not load if an Error message occurred in Add/Edit Credit Transfer


amichard:
[Confirmation Alerts | Success Message doesn't disappear |1| [v0.3.0]](https://trello.com/c/OddDUhYE/620-confirmation-alerts-success-message-doesnt-disappear-1-v030)

frontend/src/reducers/creditTransferReducer.js
```diff
@@ -77,6 +78,61 @@ const creditTransfers = (state = {
         success: true,
         items: action.data
       });
+    case 'RECEIVE_APPROVED_CREDIT_TRANSFERS':
+      return Object.assign({}, state, {
```
Sybrand: Repeat of a question I asked last week, is there a reason for doing Object.assign here, rather than 
return {...state, etc. etc.}
Am I wrong, but I don't think you're really winning anything here using Object.assign since nothing is deep copied copied anyway?

amichard: Hmmm, base reasoning is I was just following the previous code samples. I can give the code you suggested a try and see if it doesn't cause any problems.

amichard: I changed some of them to use the spread operator. I'll do more changes as I refactor the reducers more.

Sybrand: I just think it's easier to read - and less code (more elegant).

Issue #324, opened 2018-05-30
Defect | Credits Pluralization Fix |1| [v0.2.0]


amichard:
In the Text representation when viewing a credit transfer, "Credit" is not pluralized properly when the number of credits is over 999.

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Defect | Credits Pluralization Fix |1| [v0.2.0]](https://trello.com/c/UXHQKEA5/642-defect-credits-pluralization-fix-1-v020)

PR #325, opened 2018-05-30
Fixed pluralization for credits when it's over 999

amichard:
#324 

The issue was being caused by the comma when it's over 999 number of credits. So the fix is simply to do a check against numeric version of number of credits instead.

Changelog:
- Changed the condition to look into this.props.numberOfCredits, instead of this.numberOfCredits (which is a string)
- Updated condition in Visual Representation so pluralization occurs when the number of credits is over 1 (instead of 2)


amichard:
[Defect | Credits Pluralization Fix |1| [v0.2.0]](https://trello.com/c/UXHQKEA5/642-defect-credits-pluralization-fix-1-v020)

amichard:
[Defect | Credits Pluralization Fix |1| [v0.2.0]](https://trello.com/c/UXHQKEA5/642-defect-credits-pluralization-fix-1-v020)

Issue #330, opened 2018-05-31
Authorization | Add two Basic BCeID accounts to TFRS (dev & test)


amichard:
**We want to add two Basic BCeID accounts to TFRS for testing in dev and test environments.**

# Account 1

Organization: "TFRS Fantastic Fuels"

BCeID Account Details
    Registration Date: May 31, 2018
    User ID: justin.lepitzki
    Last Name: Lepitzki
    First Name: Justin
    Middle Name: 
    Other Middle Name: 
    Email: justin.lepitzki@gmail.com
    Phone: 

# Account 2 

Organization: "TFRS IMBeing Green"

BCeID Account Details
    Registration Date: May 31, 2018
    User ID: justin.wakelam
    Last Name: Wakelam
    First Name: Justin
    Middle Name: 
    Other Middle Name: 
    Email: jll7@sfu.ca
    Phone:

PR #331, opened 2018-05-31
Added 2 test fuel suppliers and 2 test users

amichard:
#330

Added fixtures to setup test basic accounts and test fuel suppliers.

Changelog:
- Added users fixture
- Added fuel suppliers fixture
- Added OrganizationManager so we don't have to know what the primary keys are for the new organizations before we can apply the fixture


amichard:
[Authorization | Add two Basic BCeID accounts to TFRS (dev & test)](https://trello.com/c/EG76eSaW/644-authorization-add-two-basic-bceid-accounts-to-tfrs-dev-test)

Issue #349, opened 2018-06-05
fix for 648-responsive-design-fix-column-width-class


rajpersram:
Changed header class as described in Trello ticket 648: https://trello.com/c/AHfuCqyH

matthewhall78:
![](https://github.trello.services/images/mini-trello-icon.png) [Responsive Design | fix column width class |1|](https://trello.com/c/AHfuCqyH/648-responsive-design-fix-column-width-class-1)

PR #350, opened 2018-06-05
changed the column class in the header to add up to 12 columns

rajpersram:
Issue #349


matthewhall78:
[Responsive Design | fix column width class |1|](https://trello.com/c/AHfuCqyH/648-responsive-design-fix-column-width-class-1)

PR #356, opened 2018-06-06
Adding the fixtures from 0.2.0 to develop

amichard:
#330 

This is to apply the same fixture that adds the test basic users and test organizations into develop


amichard:
[Authorization | Add two Basic BCeID accounts to TFRS (dev & test) |1| [0.2.0]](https://trello.com/c/EG76eSaW/644-authorization-add-two-basic-bceid-accounts-to-tfrs-dev-test-1-020)

Issue #364, opened 2018-06-11
UX Refactor | Sort Fuel Suppliers alphabetically by default for IDIR users |2|


amichard:
https://trello.com/c/GzleRArh/624-ux-refactor-sort-fuel-suppliers-alphabetically-by-default-for-idir-users-2

**Task:** 

This card is to make the Fuel Suppliers table on the Fuel Suppliers tab (for IDIR users) sort alphabetically (A to B) by Company Name by default. Currently, the Fuel Suppliers table appears to be sorting by ID# by default. 

**Demo:** 
Show that the Fuel Suppliers table sorts alphabetically by Company Name by default.

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [UX Refactor | Sort Fuel Suppliers alphabetically by default for IDIR users |2|](https://trello.com/c/GzleRArh/624-ux-refactor-sort-fuel-suppliers-alphabetically-by-default-for-idir-users-2)

PR #365, opened 2018-06-11
Changed default sorting to credits From where N/A is considered blank

amichard:
#364 

Changed the sorting method in the front-end. I did it in the front-end as we have some logic in regards to not showing from Government on PVR types. 

Changelog:
- Default sort is now Credits From, Credits To
- N/A is considered blank


amichard:
[UX Refactor | Sort Fuel Suppliers alphabetically by default for IDIR users |2|](https://trello.com/c/GzleRArh/624-ux-refactor-sort-fuel-suppliers-alphabetically-by-default-for-idir-users-2)

PR #380, opened 2018-06-15
Comments API

plasticviking:
This pull request provides a REST API to permit commenting on credit trades, with selectable visibility.

New permissions **VIEW_PRIVILEGED_COMMENTS** and **EDIT_PRIVILEGED_COMMENTS** respectively control read and write access for sensitive comments. Privileged comments are not visible to Fuel Suppliers. Unprivileged comments are (if the supplier would normally have access to that Credit Trade).

I have adjusted the fixtures to assign these permissions to existing government roles. It would also be possible in future to define government roles that do not have one or both of these permissions.

Whether a comment is sensitive or not (privileged_access flag) is an immutable property, set on creation.

I chose to implement this as a new `ViewSet` rather than as a subordinate of `CreditTrade` because no other `ViewSet` within the project requires nested routing and I wanted to avoid pulling in the `drf-nested-routes` dependency and refactoring the `CreditTrade` `ViewSet`.

A thorough test suite is included.


amichard:
Check with Kuan about updating permissions.json.

Other than that, everything looks good

plasticviking:
Split permissions into new fixture as Kuan requested

Issue #388, opened 2018-06-20
Confirmation Alerts | Rescind a Credit Transfer Proposal |1|


amichard:
https://trello.com/c/ZEBMgs4Q/384-confirmation-alerts-rescind-a-credit-transfer-proposal-1

Story: 
As a Fuel Supplier user, I want to receive a confirmation alert when I Rescind a submitted Credit Transfer Proposal so that I don't mistakenly rescind the accepted proposal. 

Scenario 1: 
GIVEN that I am a user
AND my organization is Fuel Supplier
AND I have accepted a Credit Transfer Proposal submitted to my Organization from another Fuel Supplier
AND I am viewing the accepted Credit Transfer Proposal
WHEN I click on Rescind
THEN I receive a confirmation alert that asks me to confirm that I want to rescind the submitted/accepted Credit Transfer Proposal
AND I can choose to either Rescind the offer or close the confirmation alert (i.e. cancel the action)

Scenario 2: 
GIVEN that I am a user
AND my organization is Fuel Supplier
AND I have submitted a Credit Transfer Proposal to another Fuel Supplier
AND they have not accepted that Credit Transfer Proposal
AND I am viewing the Credit Transfer Proposal screen
WHEN I click on Rescind
THEN I receive a confirmation alert that asks me to confirm that I want to rescind the Credit Transfer Proposal
AND I can choose to either Rescind the offer or close the confirmation alert (i.e. cancel the action)

Scenario 3: 
GIVEN that I am a user
AND my organization is Fuel Supplier
AND I have submitted a Credit Transfer Proposal to another Fuel Supplier
AND the other fuel supplier has accepted the Credit Transfer Proposal
AND the Director has not approved the Credit Transfer Proposal
AND I am viewing the Credit Transfer Proposal screen
WHEN I click on Rescind
THEN I receive a confirmation alert that asks me to confirm that I want to rescind the submitted/accepted Credit Transfer Proposal
AND I can choose to either Rescind the offer or close the confirmation alert (i.e. cancel the action)

Demo:
Show the confirmation alert when Rescinding a Credit Transfer Proposal under Scenario 1, 2, 3
Show that the user can either Rescind the Credit Transfer Proposal or close the confirmation alert (i.e. cancel the action)

amichard:
[...]

Issue #387, opened 2018-06-20
Credit Transfer | Add 'Rescinded' Status |5|


amichard:
https://trello.com/c/0s7CpQlS/594-credit-transfer-add-rescinded-status-5

**This card adds the Rescinded status**

**Story:**
As a Fuel Supplier user, I want to know when a Credit Transfer Proposal has been Rescinded so that I am aware that the Proposal will not be processed and approved. 

**Scenario 1:** 
GIVEN that I am a Fuel Supplier
AND I have Accepted a Credit Transfer Proposal submitted to my Organization from another Fuel Supplier (i.e. 'Sign 2/2')
AND I am viewing the Accepted Credit Transfer Proposal
WHEN I click on Rescind
AND accept the confirmation alert confirming that I want to Rescind the Proposal
THEN the Proposal is visible in the Credit Transactions History table (visible to both Fuel Suppliers & Government)
AND the status changes to Refused 

**Scenario 2:**
GIVEN that I am a Fuel Supplier
AND I have submitted a Credit Transfer Proposal to another Fuel Supplier
AND they have not accepted that Credit Transfer Proposal (e.g. 'Sign 2/2')
AND I am viewing the Credit Transfer Proposal screen
WHEN I click on Rescind
AND accept the confirmation alert confirming that I want to Rescind the Proposal
THEN the Proposal is visible in the Credit Transactions History table (visible to both Fuel Suppliers)
AND the status changes to Refused 

**Scenario 3:**
GIVEN that I am a Fuel Supplier
AND I have submitted a Credit Transfer Proposal to another Fuel Supplier
AND the other fuel supplier has accepted the Credit Transfer Proposal (i.e. 'Sign 2/2')
AND the Director has not approved the Credit Transfer Proposal
AND I am viewing the Credit Transfer Proposal screen
WHEN I click on Rescind
AND accept the confirmation alert confirming that I want to Rescind the Proposal
THEN the Proposal is visible in the Credit Transactions History table (visible to both Fuel Suppliers & Government)
AND the status changes to Refused 

**Demo:**
Show the status change when Rescinding a Credit Transfer Proposal under Scenario 1, 2, 3 (show the status change to both fuel suppliers, and Government where applicable; show that the entry is visible in the Credit Transactions table for the different organizations)

amichard:
[...]

PR #389, opened 2018-06-20
Rescinded as a Flag, instead of Status

amichard:
#387 #388 

Added the rescinded column into the credit transfer table. And use that to check instead of checking the status.

Changelog:
- Added Rescinded column to Credit Transfers Table
- Added Rescinded column to Credit Transfers History Table
- Added Rescinded to Breadcrumbs
- Minor code cleanup to improve score with pep8
- Added partialUpdateCreditTransfer that uses PATCH instead of PUT
- Added a check to allow partial updates instead of updating the whole row


amichard:
[Credit Transfer | Add 'Rescinded' Status |5|](https://trello.com/c/0s7CpQlS/594-credit-transfer-add-rescinded-status-5)

amichard:
[Confirmation Alerts | Rescind a Credit Transfer Proposal |1|](https://trello.com/c/ZEBMgs4Q/384-confirmation-alerts-rescind-a-credit-transfer-proposal-1)

amichard:
[Confirmation Alerts | Rescind a Credit Transfer Proposal |1|](https://trello.com/c/ZEBMgs4Q/384-confirmation-alerts-rescind-a-credit-transfer-proposal-1)

frontend/webpack.production.config.js
```diff
@@ -65,6 +65,7 @@ const config = {
   },
   plugins: [
     new Webpack.DefinePlugin({
+      __LOGOUT_TEST_URL__: JSON.stringify('https://logontest.gov.bc.ca/clp-cgi/logoff.cgi'),
```
plasticviking: Is there a reason to have the test URL in production config?

I think is might be better externalized to a build config file to avoid exposing this in bundle.js?

amichard:
Ya sorry. I was building the test when I pushed this up. There should now be a test for the rescinded flag which uses the patch method

frontend/webpack.production.config.js
amichard: Unfortunately, this seems to be a limitation with our deployment process. Initially, webpack.config.js was intended to be for development and webpack.production.config.js for test and prod. But it seems that dev is using production configuration as well. Kuan tried changing the settings so that dev uses webpack.config.js, but it didn't work.

Issue #412, opened 2018-06-28
History | Show Signing Authority history for a Credit Transfer Proposal |5|


amichard:
https://trello.com/c/xt33O8kK/639-history-show-signing-authority-history-for-a-credit-transfer-proposal-5

**Story:** As a TFRS user, I want to be able to see which users signed the Credit Transfer Proposal. 

**Scenario 1:**
GIVEN that I am a Gov't user
AND I have received a Credit Transfer Proposal, signed by both companies
WHEN I go to view that Proposal
THEN I am able to see the two "Signing Authority" users that signed the Proposal 

**Scenario 2:**
GIVEN that I am a Fuel Supplier user
AND I have received a Credit Transfer Proposal from another company (Signed 1/2)
WHEN I go to view the Proposal
THEN I am able to see the Signing Authority user that signed Proposal (the history entry for Sign 1/2)

**Implementation:**
The Signing Authority users from each company should appear somewhere on the Credit Transfer View Screen, perhaps in a History table? 

We want to be able to see which Signing Authority user(s) signed the Proposal at any given stage in the process. This applies to both fuel suppliers involved in the transaction and all government users. At this time, we are only concerned with the two Signing Authorities "signing" the Proposal and not any of the other history with respect to the Proposal (e.g. the user that created the proposal, the user who saved it as a draft, etc.). This information should be visible on the Credit Transfer View screen throughout the credit transfer proposal process. 

Examples:

- The respondent should be able to see which Signing Authority user signed the Proposal (Sign 1/2)
- The Government Analyst should be able to see the two Signing Authority users who signed the Proposal (Sign 1/2 and Sign 2/2) after the Proposal has been accepted (Signed 2/2).
- The Director should be able to see the two Signing Authority users who signed the Proposal (Sign 1/2 and Sign 2/2) after the Gov't Analyst has Recommended/Not Recommended the Proposal
- All parties involved (two fuel suppliers, government) should be able to see the two Signing Authority users who signed the Proposal after it has been submitted to Gov't for review (Signed 2/2) including after it has been approved/completed. 

**Demo:** 
Show the two Signing Authority users on the Credit Transfer View screen for a Credit Transfer Proposal that has been submitted to government for review
Show that this information is visible to both fuel supplier involved in the transaction, and to government users, but not to fuel suppliers not involved in the transaction.

amichard:
[...]

PR #413, opened 2018-06-28
Credit Trade Signing History

amichard:
#412

We should now have a section in the Credit Trade where it shows who signed the proposal and who reviewed the proposal

Changelog:
- Added CreditTradeSigningHistory component
- Added signatures and reviewed property in the Credit Trade
- Restored the single quotes in CreditTradeHistory (was getting into an infinite loop if I try to import CreditTradeHistory into CreditTrade)
- Removed Propose Trade permission from Government Director


amichard:
[History | Show Signing Authority history for a Credit Transfer Proposal |5|](https://trello.com/c/xt33O8kK/639-history-show-signing-authority-history-for-a-credit-transfer-proposal-5)

Issue #395, opened 2018-06-21
Credit Transfer | Fuel suppliers users can create comments |5|


dainetrinidad:
https://trello.com/c/tGlVgY8P/688-credit-transfer-fuel-suppliers-users-can-create-comments-5

PR #414, opened 2018-06-29
Add commenting for fuel suppliers

plasticviking:
This pull request introduces the commenting feature for fuel suppliers.

Changelog highlights:
- Added natural key lookups for User, Role, Permission, Credit Trade Status, and Type to ease test fixture creation
- Exhaustive testing suite to validate permissions are applied as specified by business logic
- Created a new test fixture with more production-like government roles to test finer-grained permissions
- Refactored commenting permissions to allow client and server to share the same logic (the server passes allowable commenting actions as part of a credit trade API response) 
- Tooltips/disabled statuses for action buttons when commenting mode is active


frontend/src/credit_transfers/components/CreditTransferFormButtons.js
```diff
@@ -66,22 +72,28 @@ const CreditTransferFormButtons = props => (
...
+          {Lang.BTN_REFUSE}
+        </button>
+      </TooltipWhenDisabled>
       }
       {props.actions.includes(Lang.BTN_SIGN_2_2) &&
       <TooltipWhenDisabled
-        disabled={props.disabled.BTN_SIGN_2_2}
-        title={props.permissions.BTN_SIGN_2_2
+        disabled={props.isCommenting || props.disabled.BTN_SIGN_2_2}
+        title={props.isCommenting ? Lang.TEXT_COMMENT_DIRTY : (props.permissions.BTN_SIGN_2_2
```
plasticviking: I know that eslint will complain about the nested ternary, but it's the most concise way to do this, and the parens make precedence unambiguous.

kuanfandevops:
Do you have a Trello card for this?

plasticviking:
Closes issue #395 (Trello card 688 https://trello.com/c/tGlVgY8P)

backend/api/migrations/0015_rename_comment.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
amichard: Can you adjust the filename for this so it's incremented by 1? Just to avoid two 15's

amichard:
Some notes I noticed while I was testing this:

During the review step, when I have the option to recommend and not recommend, I can't see the comment button. Am I missing some fixtures? (I can see it fine during the approval/decline step)
screen shot 2018-07-03 at 8 33 19 am

During the approval step, I noticed that there's no space between "Decline to Approve" and "Approve" buttons.
screen shot 2018-07-03 at 8 34 11 am

plasticviking:
I'll check that out. It's something to do with the spacing being different when the button is in disabled state. I haven't added any special rules so I'll dig into it.

Regarding the buttons not appearing in review state, that's a business logic rule. Per Justin analysts should be able to comment when in signed 2/2 and reviewed and directors should only be able to comment when in reviewed.

I created test fixture users for my test cases that have either RECOMMEND or (DECLINE|APPROVE) credit transfer but not both, in order to test the logic.

plasticviking:
@amichard All review comments addressed in latest push

Issue #418, opened 2018-07-05
Navigation | Create top level tab for Company Details page |1|


amichard:
**Story:**
As an external TFRS user, I want my company's information to be available separately from the FS list because this better reflects my mental model.

**Task:**
We are going to divert from the original wireframe in that we are going to split up the Fuel Suppliers List / Company Details tabs for external users.

Originally, we were going to implement a sub-menu under the 'Fuel Suppliers' tab that would include:
(1) 'List' - the publicly available list of Part 3 fuel suppliers (i.e. Market Participants list)
(2) 'Company Details' - the specific details of the organization of which the user is a part. 

Instead, we would like to have the 'Fuel Suppliers' tab take the external user directly to the 'List' webpage (Note: This was implemented under [Card 626](https://trello.com/c/1JzTr2Xe/626-fuel-suppliers-add-link-to-fuel-suppliers-list-under-fuel-suppliers-tab-for-bceid-users-1-v030)). **We then want to add a tab to the main Toolbar called 'Company Details' that will take the external user to their organization's specific details (see screenshot).**

**Additional Info:**
The Company Details tab should be placed in between the "Fuel Suppliers" tab and the "Credit Market" tab (see [Card 672](https://trello.com/c/nWvcSstg/672-credit-transfer-add-credit-market-tab-that-links-to-credit-market-report-1)). 

**Demo:**
Show the 'Company Details' tab in the main Toolbar
Show that clicking on the 'Company Details' tab in the main Toolbar takes the external user to their organization's details page.

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Navigation | Create top level tab for Company Details page |1|](https://trello.com/c/mUJOqZn4/428-navigation-create-top-level-tab-for-company-details-page-1)

PR #419, opened 2018-07-05
Added Company Details tab

amichard:
#418 #427

Company Details Tab is now available for non-government users. I created new serializers to avoid exposing more than necessary information to fuel suppliers.

Changelog:
- Added new tab for Company Details
- Company Details should contain some information, credit balance and list of users
- Added mine and member in viewsets so we don't have to fetch the organization ID first


amichard:
[Company Details | View Company Details |5|](https://trello.com/c/8BgShHzt/651-company-details-view-company-details-5)

frontend/src/constants/roles.js
```diff
@@ -0,0 +1,20 @@
+const ROLES = {
```
plasticviking: What's the rationale for not just loading this from the response?

amichard: Ah, the reason for this is that we're not necessarily using the descriptions from the backend. Such as the credit trade statuses (we have submitted and accepted in the backend, but Sign 1/2 and Sign 2/2 for the front-end).

I was debating if we should just update the description in the roles table. We can use that. But I was thinking there was a different intent originally when they designed the table. So I opt to just do a description in the front-end.

PR #431, opened 2018-07-10
Security regression

plasticviking:
This pull request creates tests for and addresses security issues discovered during the security audit.

Compliments to @amichard for security fixes, particularly with respect to credit trade status changes.

Changelog highlights:
- Legacy APIs and their routes removed
- Exhaustive tests of every possible credit trade status change for validity
- All tests reworked into common framework (and Pylint score significantly improved)
- Test that respondents cannot modify credit trade details
- Test HTTP Method access control for endpoints `/organizations`, `/compliance_periods`, and `/users` added


plasticviking:
Related Trello card #702

backend/api/serializers/CreditTrade.py
```diff
@@ -180,6 +195,18 @@ def validate(self, data):
...
+        if self.instance.respondent == request.user.organization:
+            data['fair_market_value_per_credit'] = \
+            self.instance.fair_market_value_per_credit
+
+            data['number_of_credits'] = \
+            self.instance.number_of_credits
+
+            data['zero_reason_id'] = \
+            self.instance.zero_reason_id
+
```
plasticviking: trade_effective_date, type, or compliance period?

might be easier just to load from the database and pass that as self.isntance, overwriting status?

amichard: I was considering that. But I was concerned about the is_rescinded field. Since it's not really a status field, then that also needs to be added.

But thinking more about it, it's probably better than potentially causing more issues.

Issue #439, opened 2018-07-16
History | Add Government User History to Administration Page |5|


amichard:
https://trello.com/c/hTRvY5FE/657-history-add-government-user-history-to-administration-page-5

**Story:** 
As a Government User, I want to be able to see the individual user history of everyone in my organization so that I can know who has completed which tasks and when. 

**Scenario:**
GIVEN that I am a Government User
AND I am on the Administration tab
WHEN I click on the User History sub-tab
THEN I can see the History table that shows all of the individual user history of everyone in my organization 
AND the History table is sorted by most recent entry by default

**Note:**
If the table is going to be huge, we should discuss what items to include in the table before implementing. 

**Demo:** 
Show the History table on the User History sub-tab of the Administration page
Show that completing an action in TFRS creates a History entry in the table on the User History sub-tab of the Administration page

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [History | Add Government User History to Administration Page |5|](https://trello.com/c/hTRvY5FE/657-history-add-government-user-history-to-administration-page-5)

PR #440, opened 2018-07-16
Government user history

amichard:
#439 

Added user activity in the Admin tab.

Changelog:
- Added User Activity
- Updated some fixtures
- Added additional tabs for Admin
- User Activity for government covers all users for the government as opposed to just one person


plasticviking:
Seeing a test failure. I think that some of the checks in test_credit_trade_history.py are selecting too generally (it's  picking up a trade history with status id 4 and failing, since it's asserting that a user with id 4 does not exist in the response).

amichard:
Updated the test. It shouldn't have hardcoded values. It does use the variables set in the base_test_class

amichard:
[History | Add Government User History to Administration Page |5|](https://trello.com/c/hTRvY5FE/657-history-add-government-user-history-to-administration-page-5)

Issue #442, opened 2018-07-17
Navigation | Add active class to navigation items |1|


amichard:
https://trello.com/c/qlZ0eWf6/495-navigation-add-active-class-to-navigation-items-1

**Story:**
As a TFRS Dev, I want to add a class of 'active' to active menu items (or their children) so that the CSS can highlight the active menu item. 

**Context:**
To help the CSS highlight an active menu item, we need to add a class of 'active' to menu items that are active or are children of a parent item.

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Navigation | Add active class to navigation items |1|](https://trello.com/c/qlZ0eWf6/495-navigation-add-active-class-to-navigation-items-1)

PR #443, opened 2018-07-17
Added active class on the current navigation link

amichard:
#442 

Navigation should now mark where you are now. Ran into some issues with react not updating the component unless you do a full browser refresh. So had to add a pure: false on the NavBar component. Ideally, we change this later on

Changelog:
- Changed Link to NavLink in NavBar
- NavLink should have "active" class if it matches the route


amichard:
[Navigation | Add active class to navigation items |1|](https://trello.com/c/qlZ0eWf6/495-navigation-add-active-class-to-navigation-items-1)

amichard:
[Navigation | Add active class to navigation items |1|](https://trello.com/c/qlZ0eWf6/495-navigation-add-active-class-to-navigation-items-1)

PR #449, opened 2018-07-18
Deprecate notes field

plasticviking:
Changelog:
 - Mark notes field read-only in Credit Trade serializers
 - Remove all frontend UI components referencing notes field
 - `note` can be safely deleted in a future migration once the import script has been run in all environments.


amichard:
Has Justin confirmed that we no longer need the notes field in the Historical Data Entry screen? I understand the need to deprecate it. But we should probably provide a new comment system in the historical data entry.

If Justin is okay with no comments/notes in Historical Data Entry for now, then I'm okay with this pull request.

plasticviking:
Delaying merge request until complimentary merge to include comments on historical data entry for continuity.

Issue #454, opened 2018-07-19
Security | Allow Multiple Roles for Users |8|


dainetrinidad:
We should allow multiple roles for the users. 

A lot of stuff might need be tested as currently we expect one role per user.https://trello.com/c/Vqu6m3p0/730-security-allow-multiple-roles-for-users-8

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Security | Allow Multiple Roles for Users |8| [v0.3.1]](https://trello.com/c/Vqu6m3p0/730-security-allow-multiple-roles-for-users-8-v031)

PR #456, opened 2018-07-20
Allow Multiple Roles for Users

amichard:
#454 

You should now be able to assign multiple roles to a user without the app breaking.

Changelog:
- user now has a new property called 'roles' instead of 'role'
- Merged UserMinSerializer and MemberSerializer together (They looked too similar)
- RoleSerializer no longer loads permissions
- Merged RoleSerializer and RoleMinSerializer (other than one loads permissions. they're very similar)
- User now has a permissions property (to avoid duplicate permissions through multiple roles)
- The check to see if the user is a government user has been moved from the role and into the user as a property called 'isGovernmentUser'
- Added helper function in the front-end to get the role name 
- Added test for the helper function

### DIFF
--- a/backend/api/models/User.py
+++ b/backend/api/models/User.py
@@ -31,6 +31,7 @@ from auditable.models import Auditable
 from api.managers.UserManager import UserManager
 
 from .CreditTradeHistory import CreditTradeHistory
+from .Permission import Permission
 from .Role import Role
 from .UserRole import UserRole
 
@@ -89,6 +90,18 @@ class User(AbstractUser, Auditable):
 
     def __str__(self):
         return str(self.id)
+
+    @property
+    def permissions(self):
+        """
+        Permissions that the user has based on the roles applied
+        """
+        if self.roles is None:
+            return None
+
+        return Permission.objects.distinct().filter(
+            Q(role_permissions__role__in=self.roles)
+        ).order_by('id')
 
     @property
     def roles(self):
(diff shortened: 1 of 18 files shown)


backend/api/models/User.py
```diff
@@ -117,12 +125,23 @@ def has_perm(self, permission):
         """
         Helper function to check if the user has the approrpiate permission
         """
-        if self.role is None or \
-                not self.role.permissions.filter(code=permission):
+        if self.roles is None or \
```
plasticviking: I don't think self.roles can ever be None (just an empty QuerySet)

backend/api/serializers/CreditTrade.py
```diff
@@ -402,7 +402,7 @@ def get_actions(self, obj):
 
         # If the user doesn't have any roles assigned, treat as though the user
         # doesn't have available permissions
-        if request.user.role is None:
+        if request.user.roles is None:
```
plasticviking: Again, I don't think this can happen. Should this check go away or change to a permission check if they need a specific permission to continue?

frontend/src/app/components/Navbar.js
```diff
@@ -100,8 +100,8 @@ class Navbar extends Component {
       >
         <a id="navigation-anchor" href="#navigation-anchor"><span>Navigation Bar</span></a>
         <ul className="nav navbar-nav">
-          {this.props.loggedInUser.role &&
-          this.props.loggedInUser.role.isGovernmentRole &&
+          {this.props.loggedInUser.roles &&
```
plasticviking: I'm similarly thinking that this can be simplified since it should be at worst, an empty array.

amichard:
Good catch on those. This simplified the checks better. I'll see if I can add a test.

Issue #452, opened 2018-07-19
Security | Mask 'Recommended' and 'Not Recommended' as 'Reviewed' for FS users |1|


dainetrinidad:
https://trello.com/c/ZATkJFi0/729-security-mask-recommended-and-not-recommended-as-reviewed-for-fs-users-1

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Security | Mask Recommended and Not Recommended as Reviewed for FS users |1|](https://trello.com/c/ZATkJFi0/729-security-mask-recommended-and-not-recommended-as-reviewed-for-fs-users-1)

PR #457, opened 2018-07-20
Recommended/Not Recommended should show up as Reviewed for non-government users

amichard:
#452

Recommended and Not Recommended should no longer show up in the api request and just show up as 'Reviewed'

Changelog:
- Status is now overridden in the serializer and forced to show "Reviewed" with a "Recommended" ID (regardless whether it was really recommended or not)


amichard:
[Security | Mask Recommended and Not Recommended as Reviewed for FS users |1|](https://trello.com/c/ZATkJFi0/729-security-mask-recommended-and-not-recommended-as-reviewed-for-fs-users-1)

PR #464, opened 2018-07-24
Auditing for Critical Tables

plasticviking:
Changelog:
 - Created a new migration operation AuditTable which accepts a table name and enables auditing on that table.
 - All actions logged in detail to `tfrs_audit.logged_actions`
 - Migrations are reversible.
 - Migrations are NO-OP for SQLITE/MySQL
 - Setup triggers and an audit table that log all actions modifying data for the tables specified.
 - Request HSTORE extension to Postgres, which is included with Postgres but requires superuser permission to enable. Recommend creating it manually on dev, test, and prod. eg:
 `CREATE EXTENSION HSTORE;`. This only needs to be done once.
- For developmental testing with Postgresql, since Django creates and drops the database, I recommend assigning superuser to the role used by your dev environment.

### DIFF
--- a/backend/api/migrations/0021_audit_triggers.py
+++ b/backend/api/migrations/0021_audit_triggers.py
@@ -1,0 +1,35 @@
+from django.db import migrations
+
+FILENAME = 'api/migrations/create_audit_trigger.sql'
+
+
+def create_trigger(apps, schema_editor):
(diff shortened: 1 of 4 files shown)


plasticviking:
Trello Card 731

Issue #451, opened 2018-07-18
Historical Data Entry | Replace Notes field with Comments


dainetrinidad:
https://trello.com/c/4prYEj1l/725-historical-data-entry-replace-notes-field-with-comments

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Historical Data Entry | Replace Notes field with Comments |3|](https://trello.com/c/4prYEj1l/725-historical-data-entry-replace-notes-field-with-comments-3)

PR #465, opened 2018-07-24
Replaced notes in Historical Data Admin with comments

amichard:
#451 

Replaced notes with comments in Historical Data Entry

Changelog:
- Replaced notes with a single entry comment field (cannot add more than 1 comment)
- Clean up actionTypes and reducerTypes constants
- Updated tests
- Removed some unused files


amichard:
[Historical Data Entry | Replace Notes field with Comments |3|](https://trello.com/c/4prYEj1l/725-historical-data-entry-replace-notes-field-with-comments-3)

PR #468, opened 2018-07-26
Externalize Operational Data

plasticviking:
Changelog:

- `fixtures` directory separated into operational and test categories
- `test` data is used by test cases exclusively
- `operational' data is used in dev, test, and production environments
- confidential data should not be committed to git, but can be stored in `api/fixtures/confidential` for local use
- Prodlike roles removed (and their permissions assigned to GovUser and GovDirector)
- Core data moved to migration 0023 from fixtures. This migration ___should___ be safe to run even on non-empty databases. If in doubt, take a backup. It is not reversible.
- Future core data changes (adding permissions, for example) that are not specific to the BC implementation and do not contain confidential information should be deployed via migrations so that they are versioned and reversible.
- A new Django management command, `load_ops_data` created. The command executes special operational scripts within the Django environment and logs their outcome to new table `data_load_operations`. This command can load the file either from the local system like `python manage.py load_ops_data some_file_name.py` or from Minio like `python manage.py load_ops_data --minio --minio-bucket opsdata some_file_name.py` (in which case the environment variables *MINIO_ACCESS_KEY*, *MINIO_SECRET_KEY*, and *MINIO_ENDPOINT* must be set)
- Two classes created, in api/management for data loading scripts to subclass -- `OperationalDataScript` is a generic abstract base class for complex operations and `DataLoader` is an implementation for loading lists of simple objects (in a similar way to fixtures, but reversible and tracked).


plasticviking:
Trello Card 732

kuanfandevops:
@plasticviking do you have a Trello card for this?

plasticviking:
https://trello.com/c/yoPtRXK9

Rob Johnstone

Issue #487, opened 2018-08-07
User Management | Remove inactive users ability to login to TFRS |3|


amichard:
https://trello.com/c/54WZF3OZ/737-user-management-remove-inactive-users-ability-to-login-to-tfrs-3#

**Story:** 
As a TFRS user, I want to ensure that former colleagues cannot login to TFRS so that they cannot take actions when they no longer represent my organization. 

**Scenario:** 
GIVEN that I am a TFRS user
AND my status has been set to inactive
AND I have inputted my user name and password on the TFRS login screen 
WHEN I click continue
THEN I receive an error message
AND the error message informs me that I am not able to login
AND that I should contact the User ID manager from my organization

**Demo:**
Show that an inactive user cannot login to TFRS

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [User Management | Remove inactive users ability to login to TFRS |3|](https://trello.com/c/54WZF3OZ/737-user-management-remove-inactive-users-ability-to-login-to-tfrs-3)

PR #488, opened 2018-08-07
Users that are marked as inactive can no longer login

amichard:
#487 

Inactive users should now be prevented from logging in:

Changelog:
- Additional check on authentication.py so inactive users can't login


amichard:
[User Management | Remove inactive users ability to login to TFRS |3|](https://trello.com/c/54WZF3OZ/737-user-management-remove-inactive-users-ability-to-login-to-tfrs-3)

Issue #491, opened 2018-08-07
User Management | Add list of Government users to Admin section (IDIR only) |2|


amichard:
https://trello.com/c/4NBoDQH1/739-user-management-add-list-of-government-users-to-admin-section-idir-only-2

This card is to add a list of Government users to the Administration page, similarly to how we show fuel supplier users on the Company Details page. 

**Scenario:**
GIVEN that I am a Gov't user
AND I am on the administration page
WHEN I click on the Users tab
THEN I am shown a Users table similar to what appears on the Company Details page for BCeID users

**Columns to include (same as BCeID users)**:

- Name
- Role(s)
- Email
- Phone
- Status

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [User Management | Add list of Government users to Admin section (IDIR only) |2|](https://trello.com/c/4NBoDQH1/739-user-management-add-list-of-government-users-to-admin-section-idir-only-2)

PR #492, opened 2018-08-07
Added Users Tab in Admin

amichard:
#491 

Added the Users Tab in Admin

Changelog:
- Cleaned up routes.js
- Merged the routes for Credit Transaction History, Historical Data Admin into Admin.js
- Added user routes into admin.js
- Added another route for the UserDetailsContainer


amichard:
[User Management | Add list of Government users to Admin section (IDIR only) |2|](https://trello.com/c/4NBoDQH1/739-user-management-add-list-of-government-users-to-admin-section-idir-only-2)

Issue #495, opened 2018-08-09
Credit Transfer | Allow exporting raw Fuel Supplier data to Excel-based format (IDIR users only) |8|


amichard:
https://trello.com/c/2JMF4QID/722-credit-transfer-allow-exporting-raw-fuel-supplier-data-to-excel-based-format-idir-users-only-8

**Story:** As an IDIR user, I want to be able to export Fuel Supplier company balances into an excel spreadsheet-friendly format. 

**Scenario 1:**
GIVEN that I am an IDIR user
AND I am on the Fuel Suppliers tab
WHEN I click on the export button (design TBD)
THEN I an excel-friendly file is downloaded
AND the file contains all of the relevant data for all of the Fuel Suppliers in the Fuel Suppliers table (relevant data is listed in the Information section below)

**Information:**
The following data from the Fuel Suppliers table should be exported into a spreadsheet-friendly format:

- ID
- Company Name
- Credit Balance
- Status
- Actions

Wherever possible, we should try to keep the format of the exported data the same or as similar as possible to how the data is displayed in the Fuel Suppliers table within TFRS.

**NOTE: If possible, for IDIR users, it would be great if the export function for this card and for [Card 713](https://trello.com/c/WWxN3YJ3/713-credit-transfer-allow-exporting-raw-credit-transaction-data-to-excel-based-format-idir-bceid-users) could be combined. Meaning, when I export credit transaction data ([Card 713](https://trello.com/c/WWxN3YJ3/713-credit-transfer-allow-exporting-raw-credit-transaction-data-to-excel-based-format-idir-bceid-users)) I also get the fuel supplier table data (from this card) and vice versa. **

**Demo:**

- For an IDIR user, export the Fuel Supplier data and show the excel-friendly file

Issue #494, opened 2018-08-09
Credit Transfer | Allow exporting raw credit transaction data to Excel-compatible format (IDIR & BCeID users) |8|


amichard:
https://trello.com/c/WWxN3YJ3/713-credit-transfer-allow-exporting-raw-credit-transaction-data-to-excel-compatible-format-idir-bceid-users-8

**Story:** As a TFRS user, I want to be able to export the credit transaction data into an excel spreadsheet-friendly format. 

**Scenario 1 (IDIR):**
GIVEN that I am an IDIR user
AND I am on the Credit Transactions screen
WHEN I click on the export button (design TBD)
THEN I an excel-friendly file is downloaded
AND the file contains all of the relevant data for all credit transactions in the credit transaction table (relevant data is listed in the Information section below)

**Scenario 2 (BCeID):**
GIVEN that I am a BCeID user
AND I am on the Credit Transactions screen
WHEN I click on the export button (design TBD)
THEN I an excel-friendly file is downloaded
AND the file contains all of the relevant data for my company's credit transactions in the credit transaction table (relevant data is listed in the Information section below)

NOTE: for BCeID users, the data exported should only be the credit transactions that show in their credit transactions table (the transactions that pertain to their organization) 

**Information:**
The following data from credit transactions should be exported into a spreadsheet-friendly format:

- Transaction ID
- Compliance Period
- Type
- Credits From
- Credits To
- Quantity of Credits
- Value per Credit
- Status
- Effective / Approval date (if applicable)
- Comments (**talk to PO if pulling comments becomes problematic**)

Wherever possible, we should try to keep the format of the exported data the same or as similar as possible to how the data is displayed in the Credit Transactions table within TFRS.

**Demo:**

- For an IDIR user, export the credit transaction data and show the excel-friendly file
- For a BCeID user, export the credit transaction data and show the excel-friendly file

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Credit Transfer | Allow exporting raw credit transaction data to Excel-compatible format (IDIR & BCeID users) |8|](https://trello.com/c/WWxN3YJ3/713-credit-transfer-allow-exporting-raw-credit-transaction-data-to-excel-compatible-format-idir-bceid-users-8)

PR #496, opened 2018-08-09
Export as Spreadsheet for Credit Transfers and Fuel Suppliers

amichard:
#494  #495

Added export functionality for Credit Transfers and Fuel Suppliers

Changelog:
- Created SpreadSheet Builder service
- Added Download as Spreadsheet button to Credit Transactions
- Added Download as Spreadsheet button to Fuel Suppliers

### DIFF
--- a/backend/api/services/SpreadSheetBuilder.py
+++ b/backend/api/services/SpreadSheetBuilder.py
@@ -1,0 +1,115 @@
+import xlwt
+
+
+class SpreadSheetBuilder(object):
+    """
+    Class to build spreadsheets
+    """
+    def __init__(self):
+        self.workbook = xlwt.Workbook(encoding='utf-8')
+
+    def add_credit_transfers(self, credit_trades):
+        """
+        Adds a spreadsheet for credit transfers
+        """
+        worksheet = self.workbook.add_sheet("Credit Transfers")
+        row_index = 0
+
+        columns = [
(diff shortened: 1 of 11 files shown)


amichard:
[Credit Transfer | Allow exporting raw credit transaction data to Excel-compatible format (IDIR & BCeID users) |8|](https://trello.com/c/WWxN3YJ3/713-credit-transfer-allow-exporting-raw-credit-transaction-data-to-excel-compatible-format-idir-bceid-users-8)

amichard:
[Credit Transfer | Allow exporting raw Fuel Supplier data to Excel-based format (IDIR users only) |8|](https://trello.com/c/2JMF4QID/722-credit-transfer-allow-exporting-raw-fuel-supplier-data-to-excel-based-format-idir-users-only-8)

plasticviking:
Looks good and works well. My one comment is an architectural one:

Generally, my experience is that heavy reports are handled asynchronously (request generation, pickup report as two different requests), and possibly by another server instance across a service bus, or by a scheduled single-threaded executor or similar. While I don't think we're in danger yet, generating and buffering large XLS files has the potential to be a significant resource drain on the server, especially if a bunch of them happen at once.

My opinion is that we should perhaps let this one in as-is but make a point of evaluating the technical debt associated with not developing a more robust reporting solution, especially if other reporting features are required. 

In local-testing on my machine, this version of the script allocated 150mb more memory than upstream/master during `test`.

Rob

Issue #533, opened 2018-08-17
Bug | New Credit Transfer & Credit Transaction screens repopulate values after receiving an error message |3|


amichard:
https://trello.com/c/fd1GzVZ1/767-bug-new-credit-transfer-credit-transaction-screens-repopulate-values-after-receiving-an-error-message-3

**Description:**
This bug (?) applies to both New Transfers and New Credit Transactions. 

The user creates a New Transfer (BCeID) or New Credit Transaction (IDIR), enters in the relevant data and saves a draft. When editing the saved draft, if the user removes the value(s) in a required field and then attempts to move the transaction forward (Sign 1/2 or Recommend for Approval) the system throws an error (as desired) stating that that field cannot be blank. However, at the same time as the system is throwing the error, it also re-populates the empty field with the previously deleted data. 

This card is to stop the system from re-populating the blank field. We want the user to have to re-populate this field on their own as to minimize errors in completing these tasks.

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Bug | New Credit Transfer & Credit Transaction screens repopulate values after receiving an error message |3|](https://trello.com/c/fd1GzVZ1/767-bug-new-credit-transfer-credit-transaction-screens-repopulate-values-after-receiving-an-error-message-3)

PR #534, opened 2018-08-17
Added a fix so that the fields are repopulated with the ones in the form

amichard:
#533 

Added a fix for the repopulation bug

Changelog:
- Added submitted flag to see if the user has submitted the form
- Submitted flag will check whether the fields should be populated by the current state or the original state


amichard:
[Bug | New Credit Transfer & Credit Transaction screens repopulate values after receiving an error message |3|](https://trello.com/c/fd1GzVZ1/767-bug-new-credit-transfer-credit-transaction-screens-repopulate-values-after-receiving-an-error-message-3)

Issue #539, opened 2018-08-21
History | Change transaction type in the User History tables to mirror the UI |1|


amichard:
https://trello.com/c/1DkQZyKr/781-history-change-transaction-type-in-the-user-history-tables-to-mirror-the-ui-1

**Issue:**
The Transaction Type column in the User History tables currently shows backend transaction types: Buy, Sell, Credit Validation, Part 3 Award, Credit Retirement.

**Resolution:**
Change the Transaction Type column to show transaction types as they are labelled in the frontend: Credit Transfer, Validation, Reduction, Part 3 Award

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [History | Change transaction type in the User History tables to mirror the UI |1|](https://trello.com/c/1DkQZyKr/781-history-change-transaction-type-in-the-user-history-tables-to-mirror-the-ui-1)

PR #540, opened 2018-08-21
User History Types

amichard:
#539

Updating the types for the user activity so it uses the front-end language instead.

Changelog:
- Used the getCreditTransferType function to get the front-end description for credit types


amichard:
[History | Change transaction type in the User History tables to mirror the UI |1|](https://trello.com/c/1DkQZyKr/781-history-change-transaction-type-in-the-user-history-tables-to-mirror-the-ui-1)

PR #555, opened 2018-08-30
786 - Interface Language and UI

rajpersram:
committed second revision of these files, properly altering the text of the buttons without affecting the constants.


amichard:
Please refer to #561 instead

Issue #563, opened 2018-08-30
User Management | Remove TFRS Biz test (and its users) from prod |1|


amichard:
https://trello.com/c/UPFbeTrX/753-user-management-remove-tfrs-biz-test-and-its-users-from-prod-1

We should remove the test users and organization from prod. Ok to keep in the test environment.

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [User Management | Remove TFRS Biz test (and its users) from prod |1|](https://trello.com/c/UPFbeTrX/753-user-management-remove-tfrs-biz-test-and-its-users-from-prod-1)

PR #564, opened 2018-08-30
Remove TFRS Biz Test and Users

amichard:
#563

This should delete all the credit trades, history, balance, users and organization that's associated with TFRS Biz Test

@kuanfandevops Ideally, instead of running this in DEV and TEST, we run this in a STAGING environment that's a clone of PROD. DEV and TEST both use TFRS Biz Test for example transactions. But we want to test it first so STAGING it is

### DIFF
--- a/backend/api/fixtures/operational/0011_remove_tfrs_biz_test.py
+++ b/backend/api/fixtures/operational/0011_remove_tfrs_biz_test.py
@@ -1,0 +1,88 @@
+from django.db import transaction
+from django.db.models import Q
+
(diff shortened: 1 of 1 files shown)


amichard:
[User Management | Remove TFRS Biz test (and its users) from prod |1|](https://trello.com/c/UPFbeTrX/753-user-management-remove-tfrs-biz-test-and-its-users-from-prod-1)

plasticviking:
Perhaps this shouldn't be committed to Github since it's an operational script

PR #588, opened 2018-09-12
changes for new prod pipeline

kuanfandevops:


kuanfandevops:
close without merging. a branch should be created based on tag v1.0.0 as the process is same as a hot fix.

Issue #620, opened 2018-09-20
Application | Path based URL for new pipeline |3|


amichard:
https://trello.com/c/1ExQjskc/869-application-path-based-url-for-new-pipeline-3

When apply Pull Request pipeline to TFRS, we need path based URL as we can't change the domain name. ex.
current working url is https://dev.lowcarbonfuels.gov.bc.ca/
we need the below sample url working https://dev.lowcarbonfuels.gov.bc.ca/pr-4
the path pr-4 is the pull request number

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Application | Path based URL for new pipeline |3|](https://trello.com/c/1ExQjskc/869-application-path-based-url-for-new-pipeline-3)

PR #621, opened 2018-09-20
Path Based URL for New Pipeline

amichard:
#620 

We can now mount the application in a "subfolder". 

Please make sure to update the .env file appropriately. This is shared between front-end and back-end.

.env should be in the project root folder (not in backend or frontend).

Here's a sample content:
```BASE_PATH=/pr-1```

Changelog:
- New package for NPM to allow reading of .env
- New package for django to allow reading of .env
- Updated wsgi to make the variable work

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # Visual Studio
 .vs/
+.vscode/
 
 # Django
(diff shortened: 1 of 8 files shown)


amichard:

Issue #626, opened 2018-09-20
Credit Transactions | Dual role Gov't user can see both the Pull Back and Return to Analyst buttons for a Reviewed PVR transaction |1|


amichard:
https://trello.com/c/29F4xKuh/770-credit-transactions-dual-role-govt-user-can-see-both-the-pull-back-and-return-to-analyst-buttons-for-a-reviewed-pvr-transaction

Not sure what the solution is here. Going to leave it for now. 

It will be a very rare occurrence that an IDIR user will be assigned both the Gov't Analyst and Director roles. This "defect" does not break the system and will almost never occur in reality, thus it may not be worth the effort to try and "fix" it. Creating this card just to document it as a potential issue.

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Credit Transactions | Dual role Gov't user can see both the Pull Back and Return to Analyst buttons for a Reviewed PVR transaction |1|](https://trello.com/c/29F4xKuh/770-credit-transactions-dual-role-govt-user-can-see-both-the-pull-back-and-return-to-analyst-buttons-for-a-reviewed-pvr-transaction)

PR #627, opened 2018-09-20
Show only one "Return" button when user has both analyst and director role

amichard:
#626 

Added condition to only show one button for returning the status to draft if the user has both analyst and director role.

Changelog:
- Added a check if Pull Back was already included

### DIFF
--- a/frontend/src/app/components/Navbar.js
+++ b/frontend/src/app/components/Navbar.js
@@ -197,6 +197,7 @@ class Navbar extends Component {
           }
           <li>
             <NavLink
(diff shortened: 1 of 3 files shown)


amichard:
[Credit Transactions | Dual role Gov't user can see both the Pull Back and Return to Analyst buttons for a Reviewed PVR transaction |1|](https://trello.com/c/29F4xKuh/770-credit-transactions-dual-role-govt-user-can-see-both-the-pull-back-and-return-to-analyst-buttons-for-a-reviewed-pvr-transaction)

PR #628, opened 2018-09-21
Devops pipeline notification

kuanfandevops:


kuanfandevops:
Modified three pipeline file to trigger the build of notification server and the deployment.

Issue #629, opened 2018-09-24
Add Pragma to app | Adding Pragma may mitigate caching issues with IE |1|


amichard:
https://trello.com/c/QoCnmyYT/883-add-pragma-to-app-adding-pragma-may-mitigate-caching-issues-with-ie-1

Currently IE doesn't show a refreshed list of notifications when the page is re-loaded and a refresh (F5) is required. As a user, I would like to read/unread/archive my messages and see the most up-to-date status each time I refresh the page.

This is especially important for analyst and director users who have legislative timelines to respond to supplier requests and need to be kept on top of new notifications.

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Add Pragma to app | Adding Pragma may mitigate caching issues with IE |1|](https://trello.com/c/QoCnmyYT/883-add-pragma-to-app-adding-pragma-may-mitigate-caching-issues-with-ie-1)

PR #630, opened 2018-09-24
No-Cache for Notifications

amichard:
#629

Added a never_cache decorator for IE as it seems to be running into caching issues

Changelog:
- Added never_cache decorator to list()


amichard:
[Add Pragma to app | Adding Pragma may mitigate caching issues with IE |1|](https://trello.com/c/QoCnmyYT/883-add-pragma-to-app-adding-pragma-may-mitigate-caching-issues-with-ie-1)

Issue #631, opened 2018-09-24
New Credit Transfer price field | Update price per credit field so that negative values are not permissable |2|


amichard:
https://trello.com/c/ZmUxIfOA/884-new-credit-transfer-price-field-update-price-per-credit-field-so-that-negative-values-are-not-permissable

In the New Credit Transfer page, a user is able to decrease the value in the price per credit field so the value is negative. The app then throws an error message, saying "Value per credit can't be negative." This message doesn't have to appear if the field didn't allow a negative number in the drop-down list in the first place and the user wasn't able to input a - symbol.

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [New Credit Transfer price field | Update price per credit field so that negative values are not permissable |2|](https://trello.com/c/ZmUxIfOA/884-new-credit-transfer-price-field-update-price-per-credit-field-so-that-negative-values-are-not-permissable-2)

PR #632, opened 2018-09-24
Negative values are no longer allowed for Fair Market Value Input

amichard:
#631 

Negative symbol is no longer allowed in the FairMarketValueInput

Changelog:
- Prevent keypress of -
- Paste should also detect if there's a - in the clipboard
- Tooltip should be shown notifying the user negative values are not allowed


amichard:
[New Credit Transfer price field | Update price per credit field so that negative values are not permissable |2|](https://trello.com/c/ZmUxIfOA/884-new-credit-transfer-price-field-update-price-per-credit-field-so-that-negative-values-are-not-permissable-2)

Issue #633, opened 2018-09-24
Update format of downloaded XLS files | Auto-set column width in Excel files so that user can see entire header |1|


amichard:
https://trello.com/c/B1kSP8WT/882-update-format-of-downloaded-xls-files-auto-set-column-width-in-excel-files-so-that-user-can-see-entire-header

As a user, I would like to download an Excel file of data and automatically have user-friendly column widths where I can easily see all the data. For example, the content in the rows in the Effective column displays as ####### since the column is too narrow to show the full date.

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Update format of downloaded XLS files | Auto-set column width in Excel files so that user can see entire header |1|](https://trello.com/c/B1kSP8WT/882-update-format-of-downloaded-xls-files-auto-set-column-width-in-excel-files-so-that-user-can-see-entire-header-1)

PR #634, opened 2018-09-24
Increase the width of some columns for the download spreadsheet

amichard:
#633

Increased the width of the columns for the spreadsheet so they show the full value without needing the user to manually adjust it.


amichard:
[Update format of downloaded XLS files | Auto-set column width in Excel files so that user can see entire header |1|](https://trello.com/c/B1kSP8WT/882-update-format-of-downloaded-xls-files-auto-set-column-width-in-excel-files-so-that-user-can-see-entire-header-1)

PR #636, opened 2018-09-25
KeyCloak Integration (Backend Only)

plasticviking:
Pull Request 1 of 2 related to Keycloak Integration.

# Changelog 

- Enables JWT authentication in the backend with mapping to users, Keycloak signing key retrieval, and testing.
- Keycloak is disabled by default (although tested in the test suite)
- Does not remove Siteminder (yet)
- To turn it on, set appropriate **KEYCLOAK_** environment variables for Django (see `backend/tfrs/keycloak.py` for specifics)


plasticviking:
Closing for now. I found a defect I want to address before review.

PR #663, opened 2018-10-04
Persist destination in localStorage when navigating

plasticviking:
# Changelog

- Added `targetPath` reducer that watches for **redux-oidc** action **LOAD_USER** and preserves current route in `state`.
- Added **redux-localstorage** dependency and modified store configuration to persist the state for reducer `targetPath`
- Streamlined configuration of `store` object
- minor adjustments to docker config to make it less verbose


plasticviking:
Associated with Trello card 947

Issue #662, opened 2018-10-04
User Management | Implement the Role Selection function for TFRS Admin Role |8|


amichard:
https://trello.com/c/tDWqsz9R/261-user-management-implement-the-role-selection-function-for-tfrs-admin-role-8

**Story:**
As a TFRS Admin, I want to be able to assign roles to users so that I can ensure that they have access to the parts of the system required to complete their assigned tasks. 

**Scenario:**
GIVEN that I am a TFRS User
AND my organization type is Government
AND my role is TFRS Admin
WHEN I am **adding** a user (clicking the create user button)
THEN I am able to select their role from the list of roles
AND I can select more than one role
AND the selection of the roles restricts the permissions for the user in TFRS

**Scenario:**
GIVEN that I am a TFRS User
AND my organization type is Government
AND my role is TFRS Admin
WHEN I am **updating** a user (clicking the pencil icon)
THEN I am able to select their role from the list of roles
AND I can select more than one role
AND the selection of the roles restricts the permissions for the user in TFRS

*NOTE: Screenshot may be out of date and/or missing fields*

**Demo:**
Show the list of roles when adding/editing a Government user
Show that you can select (and save) more than one role
Show that the selected role(s) actually restrict user access/capabilities based on the assigned permissions within the role(s)

PR #666, opened 2018-10-09
Added tests and allow parameter to filter the roles

amichard:
#662 

Made the roles that show up in the create/edit user more dynamic.

Changelog:
- Roles can be filtered by parameters
- Added tests

### DIFF
--- a/frontend/src/admin/users/UserAddContainer.js
+++ b/frontend/src/admin/users/UserAddContainer.js
@@ -45,7 +45,16 @@ class UserAddContainer extends Component {
 
   loadData () {
     this.props.getFuelSuppliers();
-    this.props.getRoles();
+
+    if (document.location.pathname.indexOf('/admin/') >= 0) {
+      this.props.getRoles({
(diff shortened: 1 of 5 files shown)


frontend/src/admin/users/UserAddContainer.js
```diff
@@ -45,7 +45,16 @@ class UserAddContainer extends Component {
 
   loadData () {
     this.props.getFuelSuppliers();
-    this.props.getRoles();
+
+    if (document.location.pathname.indexOf('/admin/') >= 0) {
```
plasticviking: Is it possible to choose whether to filter based on user permissions? This seems... fragile.

amichard: It does seem a bit fragile doesn't it? 

The backend does technically filter it and show you all the roles you can assign users to. So we can go that route, if ever.

But the main intention of this check is to just show the roles that are available for fuel suppliers or government users. So it's more of a logic restriction than a security check.

PR #671, opened 2018-10-10
Emailed Notifications Support

plasticviking:
Enable basic support for notification by email, per Trello card 777

### DIFF
--- a/backend/api/notifications/notifications.py
+++ b/backend/api/notifications/notifications.py
@@ -15,7 +15,7 @@ from api.models.CreditTrade import CreditTrade
 from api.models.Organization import Organization
 from api.models.Role import Role
 from api.notifications.notification_types import NotificationType
-from tfrs.settings import AMQP_CONNECTION_PARAMETERS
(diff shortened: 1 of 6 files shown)


backend/api/notifications/notifications.py
```diff
@@ -101,6 +108,30 @@ def __determine_message_recipients(is_global: bool = False,
...
+        msg['From'] = EMAIL['FROM_ADDRESS']
+        msg['To'] = email_recipient
+
+        print(msg.as_string())
+
+        with smtplib.SMTP(host=EMAIL['SMTP_SERVER_HOST'],
+                          port=EMAIL['SMTP_SERVER_PORT']) as server:
+            server.send_message(msg)
+
+        print('sending notification to {}'.format(email_recipient))
```
amichard: Is this just for us to keep track while the server is running?

plasticviking: Yes, but it might be superfluous. I'll disable the print()

Issue #672, opened 2018-10-11
Credit Transfer | Allow Fuel Supplier Users to Rescind the transfer even when the transfer is recommended |3|


amichard:
https://trello.com/c/00Vsf9sV/967-credit-transfer-allow-fuel-supplier-users-to-rescind-the-transfer-even-when-the-transfer-is-recommended-3

**Story:**
As a TFRS User, I want to be able to rescind a credit transfer even when an analyst has recommended the transfer. 

**Scenario:**
GIVEN that I am a TFRS User
WHEN I rescind a recommended credit transfer
THEN the transfer should be marked as rescinded

**Demo:**
Show that you can rescind a recommended credit transfer when you are logged in as a fuel supplier

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Credit Transfer | Allow Fuel Supplier Users to Rescind the transfer even when the transfer is recommended |3|](https://trello.com/c/00Vsf9sV/967-credit-transfer-allow-fuel-supplier-users-to-rescind-the-transfer-even-when-the-transfer-is-recommended-3)

PR #673, opened 2018-10-11
Fixed an issue where the a user cannot rescind a recommended transfer

amichard:
#672 

Fixed the issue where a fuel supplier can't rescind a recommended transfer

Changelog:
- Updated serializer so it simulates a partial update 
- Removed the reliance for front-end translation for roles


amichard:
[Credit Transfer | Allow Fuel Supplier Users to Rescind the transfer even when the transfer is recommended |3|](https://trello.com/c/00Vsf9sV/967-credit-transfer-allow-fuel-supplier-users-to-rescind-the-transfer-even-when-the-transfer-is-recommended-3)

PR #685, opened 2018-10-23
BDD-workflow tests: Historical Data Entry, Analyst Credit Transactions

NickPhura:
Additional BDD tests for Historical Data Entry and Credit Transactions as an Analyst:
- Credit Transaction
- Part 3 Award
- Validation
- Reduction

Improved how I select a row from the notifications table:
- Grabs the row with the matching text, which is unread, and which has highest transaction ID. As the tests run in order, this should guarantee I grab the correct notification each time.

Improved how the notification counts are being checked:
- Grabs the before count, and asserts it against the after count.  Additional logic to handle the fact that the notification count renders a short time after the page does.

Relocated the 'modules' folder to inside the 'pages' folder, which makes more sense.

Updated my modal logic to account for the fact that some modals return you to the current page, and others trigger a page change.

Etc.


NickPhura:
[Testing | Historical Data Entry](https://trello.com/c/W5zaqDHu/978-testing-historical-data-entry)

PR #691, opened 2018-10-25
updated db_comment for fair_market_value_per_credit

klefler:


klefler:
Ok - all changes have been made. Ready for review!

amichard:
Hi Kristin.

Most of the changes look good. I did encounter an error with the changes for Organization.py
It seems to be using tabs, which it's not liking. It prefers to use spaces instead.

Please let me know if you'd like me to push up the fix for this.

Thanks,
Richard

klefler:
Thanks Richard... I noticed there was a little error-like icon on that file... wasn't sure how to deal with it. Should I go in and fix it and resubmit?

amichard:
Yes, please. Let me know if you need help with any of these. Thanks!

klefler:
Ok. Thanks Richard. I think I have resolved the issue.

amichard:
Hi Kristin.

Unfortunately, I still encountered an error with it. The db_table_comment needs to be tabbed 1 to the left.

So from:
```
    class Meta:
        db_table = 'organization'
        db_table_comment = 'Contains a list of all of the recognized Part 3 fuel suppliers, both past and present, as well as an entry for the government which is also considered an organization.'

```
to
```
    class Meta:
        db_table = 'organization'

    db_table_comment = 'Contains a list of all of the recognized Part 3 fuel suppliers, both past and present, as well as an entry for the government which is also considered an organization.'
```

PR #692, opened 2018-10-26
1006 - add credit balance assertions

NickPhura:
Code Changes
-------------------------------------------------------------
Add credit balance assertions.

Update credit amounts (to ensure all +/- combinations of credits are unique)
Other trivial/minor updates.

Add .env support to build.gradle (for local testing against docker)
Can rename 'example.env' to '.env' and it will be picked up automatically.

**Possible Pipeline Update** 
-------------------------------------------------------------
@kuanfandevops 
We might not be setting this currently.  If not, then don't worry about this.  But if we are setting this in OpenShift we will need to update it.
- I have updated the environment variable "**BASEURL**" to now be "**BASE_URL**".

It is supposed to be BASE_URL (for consistency with other projects using BDDStack).  I just changed it before when testing something and didnt change it back.  Its not a critical change, but good to keep consistent.


NickPhura:
[Testing | Add additional credit balance assertions](https://trello.com/c/Ro701vAF/1006-testing-add-additional-credit-balance-assertions)

Issue #699, opened 2018-10-26
User Management | View the Permissions that are associated with the Fuel Supplier roles BCeID |8|


amichard:
https://trello.com/c/QXL1haUI/799-user-management-view-the-permissions-that-are-associated-with-the-fuel-supplier-roles-bceid-8

**Story:**
As a BCeID user, I want to see the permissions that are attached to the various roles so that I can understand what role I need in TFRS to complete my assigned tasks. 

**Additional Information/Context:**
We need to show the permissions that are attached to the various roles so that BCeID users can better understand what role(s) they need to be assigned. For BCeID users, we want to only show the roles that are available to Fuel Suppliers 

**Potential Design(s):**

- show the list of permissions when adding / editing a user
- show the list of permissions somewhere on the users page, or make the Role(s) clickable on the Company Details page
- Dedicated roles/permissions page? Similar to [Card 446](https://trello.com/c/MH4g5JWB/446-authorization-display-roles-and-permissions-3) but for BCeID users? Would have View only capability.

Issue #698, opened 2018-10-26
User Management | View the Permissions that are associated with Government roles IDIR |2|


amichard:
https://trello.com/c/tgS3ZxY4/937-user-management-view-the-permissions-that-are-associated-with-government-roles-idir-2

**Story:**
As an IDIR user, I want to see the permissions that are attached to the various roles so that I can understand what role I need in TFRS to complete my assigned tasks.

**Additional Information/Context:**
We need to show the permissions that are attached to the various roles so that IDIR users can better understand what role(s) they need to be assigned. For IDIR users, we want to show all roles (external and internal).

**Potential Design(s):**

- show the list of permissions when adding / editing a user
- show the list of permissions somewhere on the users page, under the Admin page
- Make the roles clickable on the Users page, under the Admin page
- Dedicated roles/permissions page? Similar to [Card 446](https://trello.com/c/MH4g5JWB/446-authorization-display-roles-and-permissions-3) but for BCeID users? Would have View only capability.

PR #700, opened 2018-10-26
Roles list

amichard:
#698 #699

Added new tab: Roles for Admin

Changelog:
- Added a new tab for Roles
- Roles should be clickable to show permissions
- Role row is clickable


amichard:
[User Management | View the Permissions that are associated with Government roles IDIR |2| v1.1.3](https://trello.com/c/tgS3ZxY4/937-user-management-view-the-permissions-that-are-associated-with-government-roles-idir-2-v113)

PR #704, opened 2018-10-31
Moved .env file into backend and frontend folders

amichard:
#620

Moved the .env files into backend and frontend folders as the pods won't have a way to go up one level.

Changelog:
- Moved where the backend is looking for the .env file into same level as manage.py
- Moved where the frontend is looking for the .env file into same level as package.json


amichard:

Issue #719, opened 2018-11-01
Notifications | Enable notifications for committed items through the Historical Data Entry |3|


amichard:
https://trello.com/c/Y3hMRrR7/997-notifications-enable-notifications-for-committed-items-through-the-historical-data-entry-3

**Task:**
This card is to enable notifications when a transaction is committed through the Historical Data Entry feature. 

Government and the fuel supplier(s) involved in the transaction should receive a notification when a transaction has been approved (committed). 

**Settings**:

- incorporate this event into the 'Approval' notifications for credit transfer proposals and PVRs

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Notifications | Enable notifications for committed items through the Historical Data Entry |3|](https://trello.com/c/Y3hMRrR7/997-notifications-enable-notifications-for-committed-items-through-the-historical-data-entry-3)

PR #720, opened 2018-11-01
Historical data entry notifications

amichard:
#719 

Notifications are now sent out when using the commit function in Historical Data Entry

Changelog:
- Added notification hook in batch process
- Related organization should be populated in the Notification Message


amichard:
[Notifications | Enable notifications for committed items through the Historical Data Entry |3|](https://trello.com/c/Y3hMRrR7/997-notifications-enable-notifications-for-committed-items-through-the-historical-data-entry-3)

PR #739, opened 2018-11-13
Fuel supplier creation

plasticviking:
# Changelog
 - Adds support for adding and editing fuel suppliers
 - Adds a "referenceData" reducer to avoid duplicating database state in constants files


plasticviking:
Trello cards 503 and 813

Issue #745, opened 2018-11-15
User Management | Restrict users' ability to log in when status is set to Inactive |1|


amichard:
https://trello.com/c/VXJNW6Ms/646-user-management-restrict-users-ability-to-log-in-when-status-is-set-to-inactive-1

**Story:**
As a TFRS PO, I want to ensure that users that are set as Inactive are not able to login to TFRS so that when a user leaves a company there is a way to remove their access. 

**Task:** 
Users with an inactive status are not able to login to TFRS

**Demo:**
- Show that an inactive user that is in the TFRS system can't login to TFRS

PR #746, opened 2018-11-15
Inactive users can no longer login

amichard:
#745 

Updated the check if the user is inactive for keycloak


amichard:
[User Management | Restrict users' ability to log in when status is set to Inactive |1|](https://trello.com/c/VXJNW6Ms/646-user-management-restrict-users-ability-to-log-in-when-status-is-set-to-inactive-1)

PR #767, opened 2018-12-05
Follow-up Fixes

amichard:
Found a couple of issues by the previous fix. So this is to fix those.

Changelog:
- Fixed logout
- If a user has multiple records of the same role, it should no longer make the checkbox unclickable
- Removed validation from Phones as it's causing confusing behaviour


amichard:
[User Management | Edit User save phone number |2|](https://trello.com/c/TKrColvH/1057-user-management-edit-user-save-phone-number)

amichard:
[User Management | User Logout |1|](https://trello.com/c/gZcAoyBJ/1056-user-management-user-logout-0)

amichard:
Ah, ya. I was thinking about that, so the migration fix should already be in place.

PR #772, opened 2018-12-07
BDD - Version updates

NickPhura:
Code Changes/Updates
-------------------------------------------------------
Update BDD build.gradle versions, to match the latest version of the BDD image built by Kuan.
Some other trivial updates.

Version
------------------------------------------------------
Rebased on latest version of 'master' as of Dec 7, 2018.

### DIFF
--- a/functional-tests/build.gradle
+++ b/functional-tests/build.gradle
@@ -13,11 +13,11 @@ ext {
     drivers = ["firefox", "firefoxHeadless", "chrome", "chromeHeadless","ie","edge","safari"]
 
     ext {
-        groovyVersion = '2.5'
+        groovyVersion = '2.5.1' // Updating this past 2.5.1 causes spock reports to fail the test execution
         gebVersion = '2.2'
         seleniumVersion = '3.141.59'
         chromeDriverVersion = '2.44'
-        geckoDriverVersion = '0.21.0'
+        geckoDriverVersion = '0.22.0'
         ieDriverVersion = '3.14.0'
         edgeDriverVersion = '3.14.0'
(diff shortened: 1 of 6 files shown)


functional-tests/build.gradle
```diff
@@ -13,11 +13,11 @@ ext {
     drivers = ["firefox", "firefoxHeadless", "chrome", "chromeHeadless","ie","edge","safari"]
 
     ext {
-        groovyVersion = '2.5'
+        groovyVersion = '2.5.0'
```
NickPhura: This should resolve the current failures.

2.5.4 exists, but we can't upgrade to it yet because spock only supports 2.5.0 at the moment.  That was an issue I found that resulted in reporting to error out (though the tests could still run).

functional-tests/build.gradle
```diff
@@ -13,11 +13,11 @@ ext {
...
         gebVersion = '2.2'
         seleniumVersion = '3.141.59'
         chromeDriverVersion = '2.44'
-        geckoDriverVersion = '0.21.0'
+        geckoDriverVersion = '0.22.0'
```
NickPhura: We aren't running other browsers yet, but I updated this anyways.

functional-tests/build.gradle
amichard: I had to use 2.5.1 on my end, but I might have a different configuration.

NickPhura: I tested 2.5.1 on my end, and everything works.  So I updated the PR with that version.

Added a comment as well, because anything pat 2.5.1 results in spock reporting to fail the test execution

PR #783, opened 2018-12-12
Fix tests

amichard:
Added delay to other parts of the tests so it doesn't fill in the updates right away.

### DIFF
--- a/functional-tests/src/test/groovy/specs/CreditTransactionSpec.groovy
+++ b/functional-tests/src/test/groovy/specs/CreditTransactionSpec.groovy
@@ -33,7 +33,7 @@ class CreditTransactionSpec extends LoggedInSpec {
   void 'Log in as an analyst and initiate a new part 3 award credit transaction'() {
     given: 'I am logged in as an Analyst'
       logInAsAnalyst()
-      Integer initialNotificationCount = headerModule.getNotificationCount(10)
+      Integer initialNotificationCount = headerModule.getNotificationCount()
     and: 'I populate all required fields for a new part 3 award'
       to NewCreditTransactionPage
       setTransactionType('Part 3 Award')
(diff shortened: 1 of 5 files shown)


functional-tests/src/test/groovy/specs/CreditTransactionSpec.groovy
```diff
@@ -51,7 +51,9 @@ class CreditTransactionSpec extends LoggedInSpec {
       at new ToastModal('Success!', 'Credit transaction recommended.')
     and: 'My unread notification count has increased by 1'
       page(HomePage)
-      headerModule.getNotificationCount() == initialNotificationCount+1
+      waitFor {
+        headerModule.getNotificationCount() == initialNotificationCount+1
```
NickPhura: HeaderModule.groovy # getNotificationCount() already has a wait built in, which defaults to 3 seconds.

In the case where the notification count does not exist (because they have no unread notifications), then a traditional waitFor{} works great, since we can wait for the element to appear and then grab the value.

But in the case where the notification count already exists, but is being updated, we can't rely on waitFor{} since it would see the element and grab it immediately, getting the old value.  So we have to resort to a fixed sleep of some number of seconds.  

So our only consistent solution is to use a fixed sleep, as in the getNotificationCount() method.

But you can pass any integer to getNotificationCount() to increase or decrease it.

functional-tests/src/test/resources/GebConfig.groovy
```diff
@@ -15,7 +15,7 @@ import org.openqa.selenium.safari.SafariDriver
 import org.openqa.selenium.remote.DesiredCapabilities
 
 waiting {
-  timeout = 20
+  timeout = 60
```
NickPhura: We should be a bit careful of increasing this too high.  

Do we really expect a single operation to take up to 60 seconds in OpenShift?
This timer isn't for the entire test to run, just each individual operation or assertion.

Though to be fair, in happy path cases, the tests will run as fast as possible and this timeout will never be hit.

In cases where tests fail, this will make the overall test run slower since it will wait for 1 minute to timeout for each failing test.

Only matters if the speed of the pipeline in failing cases matters.

functional-tests/src/test/groovy/specs/OrganizationSpecs.groovy
```diff
@@ -28,6 +28,7 @@ class OrganizationSpecs extends LoggedInSpec {
       clickCreateOrganizationButton()
       at OrganizationAddPage
     and: 'I populate all required fields to create a new organization'
+      sleep(5000)
```
NickPhura: No real suggestion here, as we already brainstormed the problem together for a while.

The one thing you could do, is turn this sleep time into an environment variable, with a default of 5 seconds, so that we can control it without making code changes.

The same thing is true for the other waits, like with the getNotificationCount() above.  Which I didn't think about until now.  Not an immediate concern, but maybe something to add later.

Issue #785, opened 2018-12-13
Secure Document Upload | file type MIME TYPE not permitted WHITELIST |1|


amichard:
https://trello.com/c/NmhvpiQw/1068-secure-document-upload-file-type-mime-type-not-permitted-whitelist-1

As a Government user, I want TFRS to not permit certain file types from being uploaded so that we can reduce our chances of receiving malicious documents. 

**Support uploading of various file types**

- doc, docx, xls, xlsx, pdf, jpg, png, txt, msg, ppt, pptx

**Not permit uploading of certain file types:**

- exe, zip, 7z, rar

PR #786, opened 2018-12-13
Added validation for the uploaded files

amichard:
#785 

Validating the file types based on mime type. The "accept" property doesn't seem to be that reliable and looks a bit awkward to use. But mimetype checking is unreliable in general.

Changelog:
- New Invalid files section
- New validation function
- onDrop sorts the files into accepted and rejected files


amichard:
Merged with #792 Please refer to that, instead.

PR #812, opened 2019-01-03
PVR Notification Messages

amichard:
#811 

Should now be more descriptive for PVR messages

Changelog:
- PVR is now being replace by the credit transaction type


amichard:
[Notifications | Notification message for PVRs |2| v1.1.3](https://trello.com/c/9jwJsdWl/982-notifications-notification-message-for-pvrs-2-v113)

PR #813, opened 2019-01-03
Clean up CreditTradeHistory, Remove Siteminder support

plasticviking:
# Changelog
 - Address Trello Card 485 by removing two superfluous DB columns from CreditTradeHistory.

# Technical Debt:
 - Remove deprecated Siteminder support in both frontend and backend
 - Remove old Siteminder headers from `User` model


amichard:
One thing to note is that you can't rollback anymore as the user_id is being set to null

Issue #818, opened 2019-01-04
User Management | Add title as an optional field when creating a new user |2|


amichard:
**Story:**
As a TFRS user, I want to see other users' titles so that I can understand the different responsibilities and jobs of users. 

**Task:**

- Add a "Title" field as an optional field when creating a new user (IDIR and BCeID)
- Show the "Title" field when viewing the user info page (individual user info page)
- Allow the "Title" field to be updated when editing a user
- Make the field a free text field

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [User Management | Add title as an optional field when creating a new user |2|](https://trello.com/c/hSK6hqt3/1096-user-management-add-title-as-an-optional-field-when-creating-a-new-user-2)

PR #819, opened 2019-01-04
Title should now be editable for Users

amichard:
#818 

Title was already in the table as a column. So I just had to update the code a bit so it's updateable

Changelog:
- Added Title in serializers
- Updated form to include title
- Updated details page to include title


amichard:
[User Management | Add title as an optional field when creating a new user |2|](https://trello.com/c/hSK6hqt3/1096-user-management-add-title-as-an-optional-field-when-creating-a-new-user-2)

Issue #831, opened 2019-01-14
Secure Document Upload | file name sanitization over 255 characters |1|


amichard:
https://trello.com/c/xjaaEPAR/1069-secure-document-upload-file-name-sanitization-over-255-characters-1

This card is to restrict uploading of documents that have a file name greater than 255 characters. This edge case is done to avoid potential issues with downloading the file to Windows, which does not allow file names greater than 255 characters.

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Secure Document Upload | Restrict file size to max 50mb |1| v1.2.7](https://trello.com/c/BHSMHD1y/1070-secure-document-upload-restrict-file-size-to-max-50mb-1-v127)

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Secure Document Upload | file name sanitization over 255 characters |1|](https://trello.com/c/xjaaEPAR/1069-secure-document-upload-file-name-sanitization-over-255-characters-1)

PR #832, opened 2019-01-14
Filename restrictions

amichard:
#831

Verified that long filenames won't cause issues with the system.

Changelog:
- Added automatic ellipsis for long filenames

### DIFF
--- a/frontend/__tests__/utils/functions.js
+++ b/frontend/__tests__/utils/functions.js
@@ -53,12 +53,12 @@ test('Validate Files should only return allowed file types', () => {
   }]);
 });
 
-test('Get File size should convert bytes to MB', () => {
+test('Get File size should convert bytes to mb', () => {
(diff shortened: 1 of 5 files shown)


amichard:
[Secure Document Upload | file name sanitization over 255 characters |1|](https://trello.com/c/xjaaEPAR/1069-secure-document-upload-file-name-sanitization-over-255-characters-1)

frontend/__tests__/utils/functions.js
```diff
@@ -53,12 +53,12 @@ test('Validate Files should only return allowed file types', () => {
   }]);
 });
 
-test('Get File size should convert bytes to MB', () => {
+test('Get File size should convert bytes to mb', () => {
```
plasticviking: the pedant in me associates a lower case b with bits, not bytes. Shouldn't it be kB?

PR #845, opened 2019-01-17
Notifications count flashing fix

plasticviking:
#Changelog
 - Track notification count in the state of `Navbar` to avoid flashing a null value while it updates


plasticviking:
Trello card 1139

PR #870, opened 2019-01-31
Celery-based orphan cleanup

plasticviking:
#Changelog
 - Added celery dependency for flexible scheduled tasks
 - Added minio orphan cleanup task

Some devops work is required to support the celery execution. I'm recommending we use supervisord as I've done in the docker installation to handle this in the openshift environments.

### DIFF
--- a/backend/api/services/DocumentService.py
+++ b/backend/api/services/DocumentService.py
@@ -65,6 +65,34 @@ class DocumentService(object):
         )
 
     @staticmethod
+    def remove_orphans():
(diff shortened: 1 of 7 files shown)


amichard:
[SFS | Create a periodic process to purge orphaned objects from the Minio bucket |3|](https://trello.com/c/IHUndoet/1142-sfs-create-a-periodic-process-to-purge-orphaned-objects-from-the-minio-bucket-3)

Issue #871, opened 2019-01-31
User Management | Creating fuel supplier user redirects view user to admin |1|


amichard:
https://trello.com/c/HT8kJvxW/1158-user-management-creating-fuel-supplier-user-redirects-view-user-to-admin-1

**Scenario:**

- IDIR user with the Administrator role
- Creating New User from an organization's company details page (Fuel Supplier > Organization > New User button)
- When I click Save User after inputting the relevant user information, TFRS saves the user and redirects me to the View User page in the Administration section 

**Issue:**

- Clicking on the Edit button allows the IDIR Administrator to assign Government roles to the Fuel Supplier user

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [User Management | Creating fuel supplier user redirects view user to admin |1|](https://trello.com/c/HT8kJvxW/1158-user-management-creating-fuel-supplier-user-redirects-view-user-to-admin-1)

PR #872, opened 2019-01-31
Fixed wrong redirect issue when creating a user

amichard:
#871

Fixed an issue where creating a fuel supplier user as a government user, redirects the user into /admin

Changelog:
- Added a check to see what the url is when the user was being created


amichard:
[User Management | Creating fuel supplier user redirects view user to admin |1|](https://trello.com/c/HT8kJvxW/1158-user-management-creating-fuel-supplier-user-redirects-view-user-to-admin-1)

PR #879, opened 2019-02-01
Mostly code clean up. And added a redirect onError in AuthCallback

amichard:
I've been trying to figure out what the issue is.

It seems like CallbackComponent might be getting called earlier than expected? Not 100% sure, but the idea of the redirect in the error function in AuthCallback is to try and get the state again. To get to that point, the user should be already authentication. The issue is that the state didn't seem to get updated.

### DIFF
--- a/frontend/src/app/App.js
+++ b/frontend/src/app/App.js
@@ -4,15 +4,11 @@ import React from 'react';
 import { connect } from 'react-redux';
 import { withRouter } from 'react-router';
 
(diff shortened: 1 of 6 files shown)


amichard:
[Compatibility | Firefox gets stuck redirecting when automapping a newly created user](https://trello.com/c/OHOFk3Ll/1171-compatibility-firefox-gets-stuck-redirecting-when-automapping-a-newly-created-user)

PR #881, opened 2019-02-04
Added metatag to force IE to use latest rendering method

amichard:
IE (even IE11) seems to be rendering the site using IE7 mode. This is causing the site not to load anything. So we have to force it to use whatever latest version it has.


amichard:
[Compatibility | IE Browser not loading TFRS Dev Test Prod](https://trello.com/c/y6ArtBAc/1170-compatibility-ie-browser-not-loading-tfrs-dev-test-prod)

PR #890, opened 2019-02-07
Fix dev

amichard:
With Rob's help, we found out what's causing the issue.

It's ReactTable updating to version 6.9.0.

I'm going to lock some versions in so we don't encounter issues like this.


amichard:
[Dev | Broken Pages | 3 |](https://trello.com/c/dZ6P8yrV/1174-dev-broken-pages-3)

Issue #898, opened 2019-02-11
SFS | Restrict the use of the archive button until IDIR user enters a record number for each file |2|


amichard:
https://trello.com/c/c49xqfpS/1117-sfs-restrict-the-use-of-the-archive-button-until-idir-user-enters-a-record-number-for-each-file-2

**Story:**
As an IDIR user, I want TFRS to require that I enter record numbers for each file submitted using the Secure File Submission feature before I can Archive the submission so that I can ensure that documents are being properly stored in TRIM before deletion from Minio. 

**Information:**
Before an IDIR user can click on the Archive button, all files within the submission must have a TRIM record number associated with them (entered by an IDIR user). 

- Field can just be plain text, we are not going to rely on TFRS to ensure valid record numbers are actually entered
- Example Record Number: D3248349F
- One record number per file submitted within a secure file submission

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [SFS | Restrict the use of the archive button until IDIR user enters a record number for each file |2|](https://trello.com/c/c49xqfpS/1117-sfs-restrict-the-use-of-the-archive-button-until-idir-user-enters-a-record-number-for-each-file-2)

Issue #897, opened 2019-02-11
SFS | Status changed to Archived after a government user clicks on Archive button |3|


amichard:
https://trello.com/c/KJ2Akb3h/1073-sfs-status-changed-to-archived-after-a-government-user-clicks-on-archive-button-3

**Task:**
This card is for changing the status of an uploaded document to Archived after a government user clicks on the Archive button. When a government user clicks on the Archive button:

- The status of that document entry changes to "Archived"
- A notification is triggered (if setting turned on)
- All documents are deleted from Minio

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [SFS | Status changed to Archived after a government user clicks on Archive button |3|](https://trello.com/c/KJ2Akb3h/1073-sfs-status-changed-to-archived-after-a-government-user-clicks-on-archive-button-3)

PR #899, opened 2019-02-11
File submission archive

amichard:
#897 #898

Added archive function for government users

Changelog:
- Archive can only be used after it's been received.
- Archive changes status to archived.
- TRIM must be provided before archiving works


amichard:
Yes. I agree. I've been lagging behind on my tests. I'll allocate time on the next sprint to make sure I create the tests.

PR #906, opened 2019-02-13
Add scan handler and celery to pipeline

kuanfandevops:


kuanfandevops:
I'd approve it by myself as it is only openshift changes and I need it in pipeline asap.

PR #918, opened 2019-02-15
Category changed to Title in SFS list

plasticviking:
# Changelog

Change 'Category' to 'Title' in SFS list
The format is {title}: {milestone} for Part 3 Evidence submissions, {title} otherwise

Trello card 1161

### DIFF
--- a/frontend/src/secure_document_upload/components/CreditTransferRequestTable.js
+++ b/frontend/src/secure_document_upload/components/CreditTransferRequestTable.js
@@ -39,9 +39,10 @@ const CreditTransferRequestTable = (props) => {
     id: 'status',
     minWidth: 50
   }, {
-    className: 'col-category',
-    Header: 'Category',
-    id: 'category',
+    accessor: item => (item.listTitle),
+    className: 'col-title',
+    Header: 'Title',
+    id: 'title',
     minWidth: 100
   }, {
     className: 'col-credit-transaction-id',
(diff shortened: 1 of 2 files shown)


amichard:
I might have some submissions that don't have milestones, but I'm running into an error when I load this with the page. 

Can we move the logic in the front-end as well? And just add a check there? Just so the back-end isn't controlled by the requirements in the front-end

plasticviking:
@amichard This one is updated too. The logic is now in the frontend and it handles the case where milestone is None.

frontend/src/secure_document_upload/components/CreditTransferRequestTable.js
```diff
@@ -39,9 +39,19 @@ const CreditTransferRequestTable = (props) => {
     id: 'status',
     minWidth: 50
   }, {
-    className: 'col-category',
-    Header: 'Category',
-    id: 'category',
+    accessor: item => {
+      if (item.type.theType === 'Evidence') {
+        if (item.milestone !== null && item.milestone.milestone.length > 0) {
```
amichard: I seem to be getting an error with the check if the document evidence is missing a milestone:
```Cannot read property 'length' of null```

The check might need to be changed to:
```if (item.milestone && item.milestone.milestone) {```
or just
```if (item.milestone) {```

plasticviking: @amichard Fixed with the latest push

PR #921, opened 2019-02-15
User First and Last name required

plasticviking:
# Changelog
 - First and Last name are required when creating or editing users
 - Error message for blank email phrasing adjusted

Trello card 1160

### DIFF
--- a/backend/api/serializers/User.py
+++ b/backend/api/serializers/User.py
@@ -22,6 +22,7 @@
 """
 from django.db.models import Q
 from rest_framework import serializers
+from rest_framework.fields import CharField
 from rest_framework.relations import PrimaryKeyRelatedField
 
 from api.models.UserCreationRequest import UserCreationRequest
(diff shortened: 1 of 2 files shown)


backend/api/serializers/User.py
```diff
@@ -111,6 +124,18 @@ class UserUpdateSerializer(serializers.ModelSerializer):
     """
     organization = OrganizationMinSerializer(read_only=True)
     roles = PrimaryKeyRelatedField(queryset=Role.objects.all(), many=True)
+    first_name = CharField(required=True, allow_blank=True)
```
amichard: Is it possible to change this into ```allow_blank=False, error_messages={"blank": "First Name is required."}``` instead?

backend/api/serializers/User.py
```diff
@@ -111,6 +124,18 @@ class UserUpdateSerializer(serializers.ModelSerializer):
...
+    last_name = CharField(required=True, allow_blank=True)
```
amichard: Same as above but with last name instead

plasticviking:
@amichard Updates finished

Issue #799, opened 2018-12-20
Data Model | Remove unnecessary columns from the CreditTradeHistory table |2|


dainetrinidad:
**Story:**
===
As a TFRS Dev, I should not see extra unnecessary audit columns for credit_trade_history so that...

CreditTradeHistory columns
===

- The initiator_id column is in the credit_trade table but not in the history table… would it make sense to add this column to the history table too?
> This should be investigated further. From the business logic, the `initiator` is populated by the `create_user`'s `organization_id`. It is a value that can never be changed once created. Adding it to the history table provides no value as of now.

- What is the purpose of the credit_trade_update_time? Is this the date the credit_trade record was updated and inserted into the history table? Would the create_timestamp column in the history table cover this?
> You are correct. The credit_history table inherits the ‘Auditable’ model we have which includes the time fields `create_timestamp`, `update_timestamp`. This was inherited codebase, so I’ll add it to our to-do list to remove this.

- What is the user_id column used for? …or is this the column you mentioned you had deleted?
> Same issue as the above. The parent model ‘Auditable' already has two user fields (`create_user` and `update_user`). A history should never be editable, so the create_user already fulfills this need to check on who made the change. Adding this to the to-do list to be removed as well.https://trello.com/c/EIgUah4w/485-data-model-remove-unnecessary-columns-from-the-credittradehistory-table-2

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Data Model | Remove unnecessary columns from the CreditTradeHistory table |2|](https://trello.com/c/EIgUah4w/485-data-model-remove-unnecessary-columns-from-the-credittradehistory-table-2)

PR #923, opened 2019-02-19
Removing Redundant columns in Credit Trade History

amichard:
#799

Removed is_internal_record, note, update_user and update_timestamp

Changelog:
- is_internal_record seems to be an old column that's no longer being used
- note has migrated to the comments table
- update_user and update_timestamp columns are redundant as they will never change from create_user and create_timestamp

@kuanfandevops before we apply this to prod. can we do a backup of our tables first? this will make changes that are irreversible


amichard:
[Data Model | Remove unnecessary columns from the CreditTradeHistory table |2|](https://trello.com/c/EIgUah4w/485-data-model-remove-unnecessary-columns-from-the-credittradehistory-table-2)

PR #930, opened 2019-02-21
Per-User Socket.IO channels

plasticviking:
# Changelog
 - This card addresses a long-standing technical debt issue. Previously, there was a single global socket.io channel and when notifications were triggered, a reload notice would distributed to all connected users.
 - This PR introduces authentication to the socket.io channel, including full JWT verification
 - notifications are now sent to specific user channels
 - Now that this channel is secure, it could easily be extended to include things like, for example, clamav scan results in realtime
 - The docker environment is suitably updated.
 - ***IMPORTANT*** the `notifications-server` pod will need `KEYCLOAK_CERTS_URL` environment defined (to match Django)


plasticviking:
Trello Card 1191 created to track this work

Issue #934, opened 2019-02-25
Fuel Codes | Add a Fuel Code |13|


amichard:
https://trello.com/c/bLjU3uem/923-fuel-codes-add-a-fuel-code-13

**Story:** 
As an IDIR user, I want to be able to input the relevant fuel code information into TFRS so that the system can track and store fuel codes. 

**Task:**

- Create a form to be filled out by an IDIR user to add a fuel code to TFRS; see "Information to input" section below
- Have 3 buttons at the bottom of the form (left to right): Back, Save Draft, Add
- Back: returns you to the previous page (effectively cancelling the adding of a fuel code)
- Save Draft: saves the fuel code information in a draft state, see [Card ###]().
- Add: adds the fuel code into TFRS

**Information to input:**

- NOTE: Number indicates tab order; brackets indicate reason why the data is collected (ACI: Approved Carbon Intensities list; CR: Compliance Reporting; I: Information purposes); Field name

- 1 [ACI,CR] Low Carbon Fuel Code (Example: BCLCF101.4, BCLCF488.11).
- 5 [ACI,CR] Effective date (calendar date; allow user to choose day from calendar)
- 6 [ACI, CR] Expiry date (calendar date; allow user to choose day from calendar)
- 7 [ACI, CR] Fuel (drop-down menu from list of approved fuels)
- 3 [ACI, CR] Carbon intensity (number value in gCO2e/MJ; restrict to two decimal places, e.g. 56.45; negative values **are** permissible)
- 2 [ACI] Company (plain text w/ autocomplete)
- 15 [ACI] Former company name (plain text w/ autocomplete)

- 4 [I] Application date (calendar date; allow user to choose day from calendar)
- 8 [I] Feedstock (plain text w/ autocomplete)
- 9 [I] Feedstock location (plain text w/ autocomplete)
- 10 [I] Feedstock miscellaneous (plain text w/ autocomplete)
- 11 [I] Fuel production facility location (plain text w/ autocomplete map API; see note)
- 12 [I] Fuel production facility nameplate capacity (number)
- 13 [I] Feedstock transport mode (pre-defined list; allow multiple selection)
- 14 [I] Finished fuel transport mode (pre-defined list; allow multiple selection)
- 16 [I] Approval date (calendar date; allow user to choose day from calendar)

**Fuel Codes:**
[...]

PR #935, opened 2019-02-25
Fuel code create

amichard:
#934 

Basic add Fuel Code should now work.

Changelog:
- new config for backend
- new config for frontend
- new migrations
- List view for Fuel Codes
- Add Fuel code
- Unit test for list and add

@kuanfandevops please note of the changes to the config. those need to go in openshift. both front end and back end need them.
/bot-ignore-length


amichard:
Sure! I'll look into adding permissions

amichard:
Permission checks added

backend/api/permissions/FuelCode.py
```diff
@@ -24,11 +24,32 @@
...
+        if request.method != 'POST' and \
+                request.user.has_perm('FUEL_CODES_VIEW'):
+            return True
+
+        return False
 
     def has_object_permission(self, request, view, obj):
-        """Check permissions When an object does exist (PUT, GET)"""
-        return True
+        """
```
plasticviking: has_permission() is called by drf before has_object_permission() -- if has_permission() returns False, has_object_permission() will not run, so this is redundant

backend/api/models/FuelCode.py
```diff
@@ -0,0 +1,126 @@
...
+    See the License for the specific language governing permissions and
+    limitations under the License.
+"""
+from django.db import models
+from auditable.models import Auditable
+
+from api.models.FuelCodeStatus import FuelCodeStatus
+
+
+class FuelCode(Auditable):
```
klefler: The column comments in general are fairly sparse. Could you please look at adding more detailed comments, giving priority to the columns that aren't as self-explanatory?

backend/api/models/FuelCode.py
```diff
@@ -0,0 +1,126 @@
...
+    """
+    Fuel codes for the Secure Document Upload
+    """
+    fuel_code = models.CharField(
```
klefler: Should this be a lookup field, or is this the latter part of a fuel code being BCLCF101.4, BCLCF488.11, etc.? Would this column contain the "BCLCF" pre-fix? In very early iterations of fuel codes, years ago, we had talked about having the BCLCF as a discreet attribute... which would allow for other types in the future.

backend/api/models/FuelCode.py
```diff
@@ -0,0 +1,126 @@
...
+        max_length=100,
+        db_comment='Low Carbon Fuel Code.'
+    )
+    company = models.CharField(
```
klefler: Would this be a free-form field or is there an opportunity to use the existing list of Organization names?

backend/api/models/FuelCode.py
```diff
@@ -0,0 +1,126 @@
...
+        max_length=100,
+        db_comment='Company.'
+    )
+    carbon_intensity = models.DecimalField(
+        blank=True,
+        null=True,
+        max_digits=999,
```
klefler: 999 seems quite large as a max length. Is there any value in decreasing the size?

backend/api/models/FuelCode.py
```diff
@@ -0,0 +1,126 @@
...
+        db_comment='Application Date.')
+    effective_date = models.DateField(
+        blank=True,
+        null=True,
+        db_comment='Effective Date.')
+    expiry_date = models.DateField(
+        blank=True,
+        null=True,
+        db_comment='Expiry Date.')
+    fuel = models.CharField(
```
klefler: Is this a list of fuel types? I.e.; Ethanol, propane, CNG? If so, this should be managed in a lookup table.

backend/api/models/FuelCode.py
```diff
@@ -0,0 +1,126 @@
...
+        blank=True,
+        null=True,
+        db_comment='Fuel.')
+    feedstock = models.CharField(
+        max_length=100,
+        blank=True,
+        null=True,
+        db_comment='Feedstock.'
+    )
+    feedstock_location = models.CharField(
```
klefler: Is this a description of a location? Or is it a lat/long... coordinates?

backend/api/models/FuelCode.py
```diff
@@ -0,0 +1,126 @@
...
+        max_length=100,
+        blank=True,
+        null=True,
+        db_comment='Feedstock Location.'
+    )
+    feedstock_misc = models.CharField(
```
klefler: Is this a comment field?

backend/api/models/FuelCode.py
```diff
@@ -0,0 +1,126 @@
...
+        max_length=100,
+        blank=True,
+        null=True,
+        db_comment='Feedstock Miscellaneous.'
+    )
+    facility_location = models.CharField(
```
klefler: Same question as feedstock_location... is this a description of a location or a coordinate?

backend/api/models/FuelCode.py
```diff
@@ -0,0 +1,126 @@
...
+        max_length=100,
+        blank=True,
+        null=True,
+        db_comment='Fuel Production Facility Location.'
+    )
+    facility_nameplate = models.IntegerField(
```
klefler: Is this a lookup value?

backend/api/models/FuelCode.py
```diff
@@ -0,0 +1,126 @@
...
+        blank=True,
+        null=True,
+        db_comment='Fuel Production Facility Nameplate Capacity.'
+    )
+    feedstock_transport_mode = models.CharField(
```
klefler: Is this a lookup value?

backend/api/models/FuelCode.py
```diff
@@ -0,0 +1,126 @@
...
+        max_length=100,
+        blank=True,
+        null=True,
+        db_comment='Feedstock Transport Mode.'
+    )
+    fuel_transport_mode = models.CharField(
```
klefler: Is this a lookup value?

backend/api/models/FuelCode.py
```diff
@@ -0,0 +1,126 @@
...
+        blank=True,
+        null=True,
+        db_comment='Former Company Name.'
+    )
+    approval_date = models.DateField(
+        blank=True,
+        null=True,
+        db_comment='Approval Date.'
+    )
+    status = models.ForeignKey(
```
klefler: Is this going to be a "status_id"? If so, the name should be updated... if it's from fuel_status_code.id

backend/api/models/FuelCode.py
```diff
@@ -0,0 +1,126 @@
...
+        FuelCodeStatus,
+        on_delete=models.PROTECT,
+        null=False
+    )
+
+    class Meta:
+        db_table = 'fuel_code'
+
+    db_table_comment = 'List of Fuel Codes .'
```
klefler: Comment should be more descriptive.

backend/api/models/FuelCodeStatus.py
```diff
@@ -0,0 +1,56 @@
...
+    limitations under the License.
+"""
+from django.db import models
+
+from api.managers.FuelCodeStatusManager import FuelCodeStatusManager
+from api.models.mixins.DisplayOrder import DisplayOrder
+from auditable.models import Auditable
+
+
+class FuelCodeStatus(Auditable, DisplayOrder):
```
klefler: Will this table have columns for description, display order, effective date, expiration date, create/update user and date? Wondering if these are created in a different part of the code?

backend/api/models/FuelCode.py
amichard: Right now it is a free-form text with the idea of BCLCF being the prefix. 

To allow more types in the future, I'll just automatically insert BCLCF as part of the string, instead of assuming it'll always be BCLCF

backend/api/models/FuelCodeStatus.py
amichard: I left out the description, effective dates and expiration dates from this table as they look unnecessary here. But they should be easy to add if it's our standard.

Create/Update user and date should be inherited from Auditable.
Same goes with Display Order

klefler: Ok - I was thinking because this table would be a code/lookup table we'd want it to be consistent with the other code/lookup tables in the model. I agree that description could be redundant because the code values are in essence the descriptions of the values... but we should consider keeping the effective/expiration as this supports showing the code value being "valid/active/etc").  And we'd expire the code rather than delete it. And because the create/update columns are on all tables, they should be here too.

That said, other options for showing "validity" of a code value, could be handled using an active_ind flag instead. It's different than the other code tables, but certainly something we can chat about.

backend/api/models/FuelCode.py
amichard: While there might be a couple of companies that's the same as fuel supplier list, most of them would be very different. Since this is more of a list of fuel producers, rather than fuel suppliers

backend/api/models/FuelCode.py
amichard: I talked to Justin about this, and it seem that 999 would be a safe number as the range can go into 3 digits.

While very rare, it's more to accommodate for it in case it does happen.

backend/api/models/FuelCode.py
amichard: It will be. I left this as a free-form for now as it'll be in a separate PR

backend/api/models/FuelCode.py
amichard: It is. I think Django just likes it named like this, but in the table itself it'll be status_id.

backend/api/models/FuelCode.py
klefler: Thanks. The way I read this was... the CI value could be up to 999 digits in length, rather than up to #999... I thought the max_digits would = 5 and decimal_places = 2. This would allow for 5 numbers in total, i.e. 999.99. Or, I could be wrong ;)

amichard: that's a good catch. let me look it up

amichard: thank you for catching that Kristin. Yes, it should be 5 not 999. I don't know which database I got that from, but that is a bit too high.

amichard:
[Fuel Codes | Add a Fuel Code |13| v1.2.16](https://trello.com/c/bLjU3uem/923-fuel-codes-add-a-fuel-code-13-v1216)

PR #944, opened 2019-02-28
SFS Status transitions

plasticviking:
# Changelog
 - Address Trello card 1192, 'Draft' SFS with file attachments no longer become 'submitted' automatically
 - Rename confusing `FSDoc` migration to `GovDoc` to keep with naming convention


backend/api/services/DocumentService.py
```diff
@@ -122,6 +122,9 @@ def validate_status(current_status, next_status):
                 next_status.status != "Draft":
             return False
 
+        if current_status.status in ["Security Scan Failed"]:
+            return False
```
amichard: This will cause Submissions with Security Scan Failed to be no longer editable.

So it can't be saved back into a draft, or remove the failing file and have it resubmitted

plasticviking: That was an intentional choice on my part since there wasn't a way to edit the files that failed anyway, however I see you've pushed a PR to enable it.

Amended to allow transition to `Draft` from `Security Scan Failed`

amichard: I think we should also allow Submitted to be a valid transition, since the user might just want to remove the file that failed and just resubmit it

plasticviking: That's not a possible transition right now, I don't think, since the only way to get to 'Submitted' is for the security scan handler to make the change on receiving the last security scan result.

We'd need to create a separate process to handle this case since the scan result handler will not be called on a file deletion.

amichard: Ah ya. That makes sense. We can deal with that later then.

Issue #940, opened 2019-02-28
Authentication | Allow multiple accounts to use the same email |8|


dainetrinidad:
**Story:**
As a BCEID user, I want to be able have multiple accounts in the system using the same email address so I don't have to manage multiple email addresses if I'm working for multiple organizations.

https://trello.com/c/tt5twM7l/1198-authentication-allow-multiple-accounts-to-use-the-same-email-8

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Authentication | Allow multiple accounts to use the same email |8|](https://trello.com/c/tt5twM7l/1198-authentication-allow-multiple-accounts-to-use-the-same-email-8)

PR #945, opened 2019-03-01
Should now allow multiple accounts to use the same email

amichard:
#940 

Users are now allowed to have multiple accounts with the same email address (As long as their BCEID and IDIR are unique)

Changelog:
- Changed keycloak_email in user_creation_request to be non-unique
- Added external_username to user_creation_request
- New unique constraint is exernal_username + keycloak_email
- Changed some names and ids in the front-end
- Updated the security scan to handle test cases
- Updated tests
- Added external username field in the front-end


amichard:
I forgot to update the actual user_creation_request portion. I'll close this for now

backend/api/keycloak_authentication.py
```diff
@@ -91,7 +96,7 @@ def authenticate(self, request):
 
         keys = self._get_keys().items()
 
-        if len(keys) == 0:
+        if not keys:
```
plasticviking: I think it is possible for `keys` to exist but be empty.

amichard: I'll revert the code

Issue #949, opened 2019-03-04
Login | Create a Login Screen |5|


amichard:
https://trello.com/c/qVyRqVSP/1210-login-create-a-login-screen-5

Instead of creating a Keycloak theme, we decided to create the login page as an alternative. Please see the attached screenshot.

We want to utilize the kc_idp_hint query parameters so we can bypass the default keycloak login. This allows us to have our customized login and make updates as needed without needing to issue a pull request to a different repository every time.

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Login | Create a Login Screen |5|](https://trello.com/c/qVyRqVSP/1210-login-create-a-login-screen-5)

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Login | Create a Login Screen |5|](https://trello.com/c/qVyRqVSP/1210-login-create-a-login-screen-5)

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Login | Create a Login Screen |5|](https://trello.com/c/qVyRqVSP/1210-login-create-a-login-screen-5)

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Login | Create a Login Screen |5|](https://trello.com/c/qVyRqVSP/1210-login-create-a-login-screen-5)

PR #950, opened 2019-03-04
Added a custom login form for TFRS

amichard:
#949 

Added a customized login form that can be enabled/disabled via config.

@kuanfandevops There will be a front-end config that needs to be added in for when this goes in dev, test and prod
```keycloak.custom_login: true```

Changelog:
- Added login page
- Added config to toggle between default redirect or custom login

Unfortunately, I can't get the login form to work unless we do some token and fetching and enable some settings in keycloak. I think the best compromise is to just use a config


amichard:
[Login | Create a Login Screen |5| v1.2.17](https://trello.com/c/qVyRqVSP/1210-login-create-a-login-screen-5-v1217)

PR #951, opened 2019-03-05
Transport mode and approved fuel selection

plasticviking:
# Changelog
 - Trello cards 939 and 940, adding fuel and feedstock transportation modes and a list of approved fuels to fuel code creation page


backend/api/models/ApprovedFuel.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def natural_key(self):
+        """
+        Allows type 'name' to be used to identify
+        a row in the table
+        """
+        return (self.name,)
+
+    class Meta:
+        db_table = 'approved_fuel'
```
klefler: Please consider changing the name of this table to approved_fuel_type. This will help identify it as a lookup table and will keep naming consistent with others within the model.

backend/api/models/ApprovedFuel.py
```diff
@@ -0,0 +1,54 @@
...
+
+    db_table_comment = "List of approved fuel types"
```
klefler: If possible, could you add a little more to the table comment to help describe what types of values would be in this table? For example; Contains a list of approved fuel types such as CNG, Electricity, Ethanol, LNG, Propane, etc.

backend/api/models/TransportMode.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def natural_key(self):
+        """
+        Allows type 'name' to be used to identify
+        a row in the table
+        """
+        return (self.name,)
+
+    class Meta:
+        db_table = 'fuel_transport_mode'
```
klefler: Please consider changing the name of this table to fuel_transport_mode_type or fuel_transport_type. This will help identify it as a lookup table and will keep naming consistent with others within the model.

backend/api/models/TransportMode.py
```diff
@@ -0,0 +1,94 @@
...
+
+    fuel_code = models.ForeignKey(
+        'FuelCode',
+        on_delete=models.PROTECT
+    )
+
+    class Meta:
+        db_table = 'feedstock_transport_mode_fuel_code'
+
+    db_table_comment = "Relationship table for fuel codes to feedstock transport modes"
```
klefler: I'm trying to think if this needs some re-wording as I'm wondering if someone could easily interpret the meaning of this table. I don't want to impose a big change... so would it be ok to update the description? Something like "Contains a list of transportation modes used to transport feedstock for a specific fuel code."

backend/api/models/TransportMode.py
```diff
@@ -0,0 +1,94 @@
...
+
+    fuel_code = models.ForeignKey(
+        'FuelCode',
+        on_delete=models.PROTECT
+    )
+
+    class Meta:
+        db_table = 'fuel_transport_mode_fuel_code'
+
+    db_table_comment = "Relationship table for fuel codes to finished fuel transport modes"
```
klefler: Same thoughts as the other xref table above... could we change the description to:
"Contains a list of transportation modes used to transport finished fuel for a specific fuel code."

plasticviking:
@klefler I've addressed your comments with the latest push

klefler:
@plasticviking Thanks! Looks great!

Issue #953, opened 2019-03-05
SFS | Allow IDIR user to add comments to submissions with a Received status |2|


amichard:
https://trello.com/c/ROew5uBs/1194-sfs-allow-idir-user-to-add-comments-to-submissions-with-a-received-status-2

**Story:**
As an IDIR user with the File Submission (Government) role, I want to be able to add internal and external comments when the submission is in the Received status. 

- Currently, an IDIR user is only able to add comments (internal and external) when the status of the submission is Submitted. This card is to add the ability for the IDIR user to create comments (internal and external) when the status is Received. 

**Note:**
Comments should still be restricted (i.e. not allowed) when the submission has an Archived status.

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [SFS | Allow IDIR user to add comments to submissions with a Received status |2|](https://trello.com/c/ROew5uBs/1194-sfs-allow-idir-user-to-add-comments-to-submissions-with-a-received-status-2)

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [SFS | Adjust error message when trying to rescind a submission that has just been marked as received |1|](https://trello.com/c/CCAh7sqv/1193-sfs-adjust-error-message-when-trying-to-rescind-a-submission-that-has-just-been-marked-as-received-1)

Issue #952, opened 2019-03-05
SFS | Adjust error message when trying to rescind a submission that has just been marked as received |1|


amichard:
https://trello.com/c/CCAh7sqv/1193-sfs-adjust-error-message-when-trying-to-rescind-a-submission-that-has-just-been-marked-as-received-1

**Background Scenario:**

- A fuel supplier BCeID user with the File Submission role is viewing the details of a submitted SFS submission
- At the same time, an IDIR user marks the submission as Received
- While still on the view submission page, the BCeID user attempts to rescind the submission. This causes an error message to appear (as intended). 

This card is to adjust the language of the error message. 

**Current messaging:**
"Error! Submission cannot be Draft as it currently has a status of Received."

**Revised messaging:**
"Error! The submission cannot be rescinded because it has been marked as received by a Government user. Please refresh your browser."

PR #954, opened 2019-03-05
Government reviewers can still add comments on received submissions

amichard:
#952 #953 

Updated the flow so Government reviewers can still add comments on received submissions.
Also, updated the error message when trying to rescind a submission that's been received.

Changelog:
- Changed so that we check with the backend, instead of the front-end on whether to show the comment buttons
- Updated the check in the permission so that it checks whether the status is in the allowed statuses, instead of checking whether the status is forbidden
- During validation, we no longer check for the status directly. Instead, we check if the user is authorized to make that change (which checks the status internally)

### DIFF
--- a/backend/api/permissions/DocumentComment.py
+++ b/backend/api/permissions/DocumentComment.py
@@ -34,16 +34,17 @@ class DocumentCommentPermissions(permissions.BasePermission):
         """
         Check whether the user should have authority to add a comment.
         Government Users with abilities to review the documents should
-        always have authority to add a comment.
+        always have authority to add a comment, unless it's archived.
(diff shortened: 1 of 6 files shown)


amichard:
[SFS | Adjust error message when trying to rescind a submission that has just been marked as received |1|](https://trello.com/c/CCAh7sqv/1193-sfs-adjust-error-message-when-trying-to-rescind-a-submission-that-has-just-been-marked-as-received-1)

amichard:
[SFS | Allow IDIR user to add comments to submissions with a Received status |2|](https://trello.com/c/ROew5uBs/1194-sfs-allow-idir-user-to-add-comments-to-submissions-with-a-received-status-2)

Issue #959, opened 2019-03-05
Notifications | Inactive users don't receive notifications |1|


amichard:
https://trello.com/c/ghN5HCdo/1202-notifications-inactive-users-dont-receive-notifications-1

**Story:**
As a TFRS user, I don't want to receive email notifications after I have left the organization (when my TFRS status is set to inactive).

**Issue:**
Currently when a user will still receive email notifications when they are set to inactive. Since an inactive status prevents the user from logging in to TFRS, they are unable to change their notification settings (stop the emails). 

**Proposed Solution:**
Have the email notification system check if the user is active before sending the notification.

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Notifications | Inactive users don't receive notifications |1|](https://trello.com/c/ghN5HCdo/1202-notifications-inactive-users-dont-receive-notifications-1)

PR #960, opened 2019-03-05
Notifications will no longer be sent to inactive users.

amichard:
#959 

Added a check so that notifications are no longer sent to inactive users.

Changelog:
- Added a check so that recipients that are inactive are no longer sent notifications
- Did some code clean up as well


amichard:
[Notifications | Inactive users don't receive notifications |1|](https://trello.com/c/ghN5HCdo/1202-notifications-inactive-users-dont-receive-notifications-1)

PR #961, opened 2019-03-06
Autocomplete

plasticviking:
# Changelog
 - Five fields on the fuel code creation page are now autocompleted with server-side values per Trello card 942
 - *IMPORTANT* Postgresql needs `postgresql-contrib` extension `pg_trgm` installed and available or migration 91 will fail (this is similar to what was needed for the `hstore` extension and may require permissions changes or a manual intervention to create the extension)
 - added environment variable `FIX_BROKEN_SQLITE`, which if set to True, will allow Django 1.11 to work with versions of Sqlite >= 3.26


amichard:
screen shot 2019-03-06 at 8 09 42 am

I'm noticing the autocomplete fields no longer have that blocked, and full width class. Can you please apply them?

amichard:
I'm also having a hard time trying to get the autocomplete to work.

Is it adding a fuel code, then it should try to autofill next time?

plasticviking:
@amichard the latest push addresses the styling changes and brings them into alignment with the other fields

kuanfandevops:
@plasticviking there are conflicts frontend/src/admin/fuel_codes/FuelCodeAddContainer.js

PR #968, opened 2019-03-07
Removed the extra api at the end

amichard:
#949 

I misread what BASE_URL will do. So now it's adding /api at the end. This PR should fix that by just finding out where the user currently is.


amichard:
[Login | Create a Login Screen |5| v1.2.17](https://trello.com/c/qVyRqVSP/1210-login-create-a-login-screen-5-v1217)

Issue #991, opened 2019-03-14
Notifications | IDIR users don't receive a notification when a SFS is submitted |1|


amichard:
https://trello.com/c/n8b85CPY/1261-notifications-idir-users-dont-receive-a-notification-when-a-sfs-is-submitted-1

**Bug Description:**
IDIR users do not receive a notification (in-app or email) when a SFS is submitted by a fuel supplier ("File Submitted" action) despite having the notification turned on in the Settings

**Desired state:**
IDIR users receive notifications, both in-app and email, when a SFS is submitted by a fuel supplier and they have the action turned on in the notification settings

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Notifications | IDIR users don't receive a notification when a SFS is submitted |1|](https://trello.com/c/n8b85CPY/1261-notifications-idir-users-dont-receive-a-notification-when-a-sfs-is-submitted-1)

PR #992, opened 2019-03-14
Submit notification is now sent to government users

amichard:
#991

Submitted Secure File Submissions Notifications should be sent to government users now

Changelog:
- Changed to use the helper function to send the notifications
- Additional check for files that are marked failed


amichard:
[Notifications | IDIR users don't receive a notification when a SFS is submitted |1|](https://trello.com/c/n8b85CPY/1261-notifications-idir-users-dont-receive-a-notification-when-a-sfs-is-submitted-1)

Issue #1003, opened 2019-03-15
Fuel Codes | Autocomplete must be lowercase |1|


amichard:
https://trello.com/c/RY6NySMF/1259-fuel-codes-autocomplete-must-be-lowercase-1

**Location:**
IDIR user when adding or updating a fuel code, Administration section > Fuel Codes > "Fuel" drop-down menu

**Bug Description:**
The autocomplete function only works when you are inputting text in the autocomplete fields in lowercase. Inputting data in uppercase format breaks the autocomplete functionality. 

**Desired state:**
An autocomplete function that is not case sensitive and works in all cases (upper, lower, some combination of both)

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Fuel Codes | Autocomplete must be lowercase |1|](https://trello.com/c/RY6NySMF/1259-fuel-codes-autocomplete-must-be-lowercase-1)

Issue #1002, opened 2019-03-15
Fuel Codes | Unable to use default fuel type selection without selecting another fuel first then re-selecting default |1|


amichard:
https://trello.com/c/xU3MkDK6/1258-fuel-codes-unable-to-use-default-fuel-type-selection-without-selecting-another-fuel-first-then-re-selecting-default-1

**Location:**
IDIR user when adding or updating a fuel code, Administration section > Fuel Codes > "Fuel" drop-down menu

**Bug Description:**
The Fuel drop-down menu defaults to the first fuel in the list (Biodiesel). There is a bug that makes you select a fuel type before proceeding, even if the fuel in question is the default fuel (biodiesel). In other words, if you are adding or updating a biodiesel fuel code, you must select another fuel from the list and then re-select biodiesel; you are not able to just leave the fuel as the default selection of biodiesel. 

**Desired state:**
An IDIR user can leave the default selection of biodiesel when adding or updating a fuel code

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Fuel Codes | Unable to use default fuel type selection without selecting another fuel first then re-selecting default |1|](https://trello.com/c/xU3MkDK6/1258-fuel-codes-unable-to-use-default-fuel-type-selection-without-selecting-another-fuel-first-then-re-selecting-default-1)

Issue #1001, opened 2019-03-15
Fuel Codes | Adjust fuel production nameplate label when adding or updating a fuel code |1|


amichard:
https://trello.com/c/6EJ7mIbg/1255-fuel-codes-adjust-fuel-production-nameplate-label-when-adding-or-updating-a-fuel-code-1

**Background Info:**

- IDIR user
- New fuel code page (Administration tab > Fuel Codes > New Fuel Code)
- When adding or updating a fuel code

**Current label:**
Fuel Production Facility Nameplate Capacity:

**Revised label:**
Fuel Production Facility Nameplate Capacity (litres/GJ per year):

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Fuel Codes | Adjust fuel production nameplate label when adding or updating a fuel code |1|](https://trello.com/c/6EJ7mIbg/1255-fuel-codes-adjust-fuel-production-nameplate-label-when-adding-or-updating-a-fuel-code-1)

Issue #1000, opened 2019-03-15
Fuel Codes | Adjust Feedstock and Finished fuel transport labels when adding or updating a fuel code |1|


amichard:
https://trello.com/c/YtFHQUIH/1257-fuel-codes-adjust-feedstock-and-finished-fuel-transport-labels-when-adding-or-updating-a-fuel-code-1

**Background Info:**

- IDIR user
- New fuel code page (Administration tab > Fuel Codes > New Fuel Code)
- When adding or updating a fuel code

**Current labels:**

- Feedstock Transport Mode:
- Finished Fuel Transport Mode: 

**Revised label:**

- Feedstock Transport Mode (use ctrl key to select multiple modes):
- Finished Fuel Transport Mode: 
Fuel Production Facility Nameplate Capacity (use ctrl key to select multiple modes):

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Fuel Codes | Adjust Feedstock and Finished fuel transport labels when adding or updating a fuel code |1|](https://trello.com/c/YtFHQUIH/1257-fuel-codes-adjust-feedstock-and-finished-fuel-transport-labels-when-adding-or-updating-a-fuel-code-1)

PR #1004, opened 2019-03-15
Fuel code ui changes

amichard:
#999 #1000 #1001 #1002 #1003 

Label changes to help the user understand the format for the input fields when adding a fuel code

Changelog:
- Label changes
- Added a blank default for fuel
- forced lower case when doing autocomplete

Note: I'm still running into some issues with the autocomplete even with the enforce lowercase. After a couple more trials, it might be a caching issue as it doesn't seem to detect the new company until after you close and re-open the browser.

### DIFF
--- a/backend/api/services/Autocomplete.py
+++ b/backend/api/services/Autocomplete.py
@@ -77,17 +77,22 @@ class SimpleDatabaseCompletion(Completion):
         elif connection.vendor == 'postgresql':
             return self._get_matches_postgres(q)
         else:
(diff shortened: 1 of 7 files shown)


amichard:
[Fuel Codes | Selecting the Fuel Codes sub tab causes the sub tab menu to disappear |0|](https://trello.com/c/ctWhbkMO/1287-fuel-codes-selecting-the-fuel-codes-sub-tab-causes-the-sub-tab-menu-to-disappear-0)

PR #1023, opened 2019-03-21
Changed the validation slightly so that it has a different check with…

amichard:
#1022 

Added Draft Validation

Changelog:
- Split the validation messages into two, for drafts and approval


amichard:
[Fuel Codes | Set certain fields as required when Saving a Draft fuel code |2|](https://trello.com/c/ln268Bgz/954-fuel-codes-set-certain-fields-as-required-when-saving-a-draft-fuel-code-2)

PR #1034, opened 2019-03-25
Frontend Optimizations

plasticviking:
# Changelog
- Facility Location is now autocomplete-enabled
- Notifications will once again live-reload (but only if you're on the notifications page)
- Autocomplete cache invalidation on certain actions keeps autocompleted recommendations fresh


kuanfandevops:
which trello card is associated with this PR? @plasticviking

Issue #1035, opened 2019-03-25
Notifications | Hide certain Notifications in Settings for BCeID users |0|


amichard:
https://trello.com/c/rCWxzlTH/1294-notifications-hide-certain-notifications-in-settings-for-bceid-users-0

**Background:**
Currently, the notification settings page shows notification options for features that have not been released.

**Task:**
To avoid potential confusion, **for BCeID users**, only show the notification settings for features that are currently released in TFRS. 

**NOTE:** We will want to show the additional notification settings upon release of the various features (credit transfers, file submission, etc.). 

**Settings to Show:**

- Credit Transfer Proposals: Director Approval, Director Declined to Approve
- PVRs: Director Approval, Director Declined to Approve 
- File Submission: None

**Settings to Hide:**

- Credit Transfer Proposals: Draft Saved, Signed 1/2, Refused, Signed 2/2, Rescinded
- PVRs: None
- File Submission: All of them (File Pending Submission, File Submitted, Security Scan Failed, Submission Received, Submission Archived)

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Notifications | Hide certain Notifications in Settings for BCeID users |0|](https://trello.com/c/rCWxzlTH/1294-notifications-hide-certain-notifications-in-settings-for-bceid-users-0)

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Notifications | Hide certain Notifications in Settings for BCeID users |0|](https://trello.com/c/rCWxzlTH/1294-notifications-hide-certain-notifications-in-settings-for-bceid-users-0)

PR #1036, opened 2019-03-25
Notification Settings Update

amichard:
#1035

Changelog:
- Notification settings for Secure File Submission will now only show up if the feature is enabled
- Some Notification settings for Credit Transfer only show up when the new config credit_transfer.enabled is set to true

@kuanfandevops we need a new setting for the front-end:
```
credit_transfer.enabled: true
```

Please add this to DEV and TEST, but not in PROD

### DIFF
--- a/frontend/src/settings/components/SettingsDetails.js
+++ b/frontend/src/settings/components/SettingsDetails.js
@@ -40,11 +40,23 @@ const SettingsDetails = props => (
         <NotificationsCreditTransactionsTable
           addToFields={props.addToFields}
           fields={props.fields.settings.notifications}
-          items={CREDIT_TRANSFER_NOTIFICATIONS.filter(notification =>
-            (props.loggedInUser.isGovernmentUser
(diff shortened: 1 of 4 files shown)


frontend/src/settings/components/SettingsDetails.js
```diff
@@ -62,9 +75,12 @@ const SettingsDetails = props => (
...
           toggleCheck={props.toggleCheck}
           type="government-transfer"
-        />,
+        />
+      ]}
+      {!props.subscriptions.isFetching && props.subscriptions.success &&
+      CONFIG.SECURE_DOCUMENT_UPLOAD.ENABLED && [
         <h3 key="header-doc">
-          File Submission
+          File Submission {CONFIG.SECURE_DOCUMENT_UPLOAD.ENABLED}
```
plasticviking: Why display the boolean here?

amichard: nice catch. whoops

Issue #1052, opened 2019-03-28
Tables | Remove the open box icon column on certain TFRS tables |1|


amichard:
https://trello.com/c/8FfGcG98/1330-tables-remove-the-open-box-icon-column-on-certain-tfrs-tables-1

**Task:**
Remove the open/box icon column on the tables listed below. Since we have made the entry rows within these tables clickable, the open box icon column on the far right hand-side of the tables are no longer necessary. 

**Affected Tables:** 

- Credit Transactions table (BCeID & IDIR)
- Users table (BCeID & IDIR)
- Roles table (BCeID & IDIR)
- User Activity table (BCeID & IDIR)
- Secure File Submissions table (BCeID & IDIR)
- Fuel Codes table (IDIR)
- Other?

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Tables | Remove the open box icon column on certain TFRS tables |1|](https://trello.com/c/8FfGcG98/1330-tables-remove-the-open-box-icon-column-on-certain-tfrs-tables-1)

PR #1053, opened 2019-03-28
Removed the open box icon

amichard:
#1052 

The rows are clickable, and the box open isn't clear on what kind of action it'll do. So we opt to just get rid of it.

Changelog:
- Removed open box column
- Removed magnifying glass (it was just for display anyway


kuanfandevops:
@amichard there is conflict in PR 1053

Issue #1054, opened 2019-03-28
Fuel Codes | Expiry date must be after Effective date |1|


amichard:
https://trello.com/c/DcXtkdnE/1325-fuel-codes-expiry-date-must-be-after-effective-date-1

**Task:**
Implement a validation check that ensures that the date of expiry (expiry date field) is after the date the code comes into effect (effective date field).

**Purpose:**
An added check to make sure that the effective dates are entered correctly and the gov't analyst did not make a mistake

**Error Message:**
"Error! The expiry date proceeds the effective date."

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Fuel Codes | Expiry date must be after Effective date |1|](https://trello.com/c/DcXtkdnE/1325-fuel-codes-expiry-date-must-be-after-effective-date-1)

PR #1055, opened 2019-03-28
Added a check in both front-end and back-end for the expiry and effective dates

amichard:
#1054

Added an additional validation to make sure that the expiry date is always after the effective date.

Changelog:
- Added a front-end check to make sure the expiry date is after the effective date
- Added a back-end check to make sure the expiry date is after the effective date when creating or updating


amichard:
[Fuel Codes | Expiry date must be after Effective date |1|](https://trello.com/c/DcXtkdnE/1325-fuel-codes-expiry-date-must-be-after-effective-date-1)

amichard:
Thanks, Rob!

Issue #1064, opened 2019-04-01
Fuel Codes | Adjustment to sorting the Low Carbon Fuel Code column ascending descending |1|


dainetrinidad:
**Location:**
View Fuel Code table (IDIR only)

- Administration tab
- Fuel Codes sub-tab
- Column: Low Carbon Fuel Code

**Background:**
A unique fuel pathway is given a single fuel code (e.g. BCLCF101.0); the number after the decimal illustrates a new version of the same fuel pathway. For example, when BCLCF101.0 expires the fuel producer will apply for renewal. The updated pathway will be given a new fuel code (BCLCF101.1) and a new entry created in the Fuel Code table. 

When sorting these values by ascending/descending, TFRS should account for the number(s) after the decimal in determining which value is "higher" or "lower". 

**Examples:**

- BCLCF102.0 is "larger" than BCLCF101.0, which is "larger" than BCLCF101.5
- BCLCF105.10 is "larger" than BCLCF105.5, which is larger than BCLCF105.1 

**Task:**
When sorting the Low Carbon Fuel Code column by descending/ascending order (as is the default for the Fuel Codes sub-tab page), have the system account for the number(s) after the decimal in determining the appropriate order, as described above. https://trello.com/c/Uddr72G2/1320-fuel-codes-adjustment-to-sorting-the-low-carbon-fuel-code-column-ascending-descending-1

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Fuel Codes | Adjustment to sorting the Low Carbon Fuel Code column ascending descending |1|](https://trello.com/c/Uddr72G2/1320-fuel-codes-adjustment-to-sorting-the-low-carbon-fuel-code-column-ascending-descending-1)

Issue #1063, opened 2019-04-01
Fuel Codes | Autocomplete next version of Low Carbon Fuel Code |5|


dainetrinidad:
**Task:** 
As an IDIR user, I want TFRS to autocomplete the next version of the Low Carbon Fuel Code field when adding a new fuel code.

**Scenario:**
GIVEN that I am an IDIR user
AND I am adding a fuel code
WHEN I select the Low Carbon Fuel Code field 
AND I start typing in the field
THEN TFRS autocomplete the next available version of the fuel code
AND I am unable to choose a different version
AND I am still able to modify the code (just not the version)

**Edge Case Error Message:**

- If two analysts are entering in the same new fuel code at the same time, the error message should read: "Error! The fuel code BCLCFXXX.X is already in use. Please consult the Fuel Codes table to ensure that this entry does not already exist." 
- In this case, BCLCFXXX.X is the specific code in question. https://trello.com/c/vMBkekOK/1319-fuel-codes-autocomplete-next-version-of-low-carbon-fuel-code-5

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Fuel Codes | Autocomplete next version of Low Carbon Fuel Code |5|](https://trello.com/c/vMBkekOK/1319-fuel-codes-autocomplete-next-version-of-low-carbon-fuel-code-5)

PR #1065, opened 2019-04-01
Fuel code split

amichard:
#1063 #1064 

As discussed during our sprint planning, we're going to separate the fuel code into three parts

Changelog:
- Added migration to separate the fuel code, version and minor version
- Added auto-complete functionality to show the next available version
- Added test for the auto-complete functionality
- Auto-complete functionality automatically removes non-numeric characters
- Fixed sorting so that the minor version is counted as separate, instead of as a decimal


backend/api/serializers/FuelCode.py
```diff
@@ -100,6 +103,52 @@ class FuelCodeSaveSerializer(serializers.ModelSerializer):
...
+        fuel_code_version = data.get('fuel_code_version')
+        fuel_code_version_minor = data.get('fuel_code_version_minor')
+
+        matches = FuelCode.objects.filter(
+            fuel_code=fuel_code,
+            fuel_code_version=fuel_code_version,
+            fuel_code_version_minor=fuel_code_version_minor
+        )
+
+        print(fuel_code)
```
plasticviking: For debugging?

backend/api/serializers/FuelCode.py
```diff
@@ -100,6 +103,52 @@ class FuelCodeSaveSerializer(serializers.ModelSerializer):
...
+                    fuel_code_version,
+                    fuel_code_version_minor
+                )
+            )
+
+        next_available_version = Autocomplete.get_matches(
+            'fuel_code.fuel_code_version', str(fuel_code_version))
+
+        print(next_available_version)
+        print('wow')
```
plasticviking: More debugging?

backend/api/serializers/FuelCode.py
amichard: whoops. thanks for catching this. ill get rid of it while i fix the merge conflict

Issue #1091, opened 2019-04-08
Credit Calculation | View Energy Densities |5|


dainetrinidad:
**Story:**
As an IDIR user, I want to be able to view the energy densities for the different fuels so that I can understand how current and future credits will be calculated. 

- This card is to build a table in both the back-end and front-end to track and view the energy density values used in the calculation of credits (section 6 of the Act).
- The table can be found in section 11.02 (3) of the [Regulation](http://www.bclaws.ca/EPLibraries/bclaws_new/document/ID/freeside/394_2008). Screenshot attached.

**Scenario 1:**
GIVEN that I am an IDIR user
AND I am on the Administration tab
WHEN I click on the "Compliance Reporting" sub-tab
THEN I am taken to the Compliance Reporting sub-tab within the Administration tab
AND I can view the Energy Densities

**Forward looking usage:**

- We will be using/pulling these values when we implement compliance reporting as they are used in the calculation of credits/debits. 
- We will want IDIR users to be able to edit these values as they change from time to time.https://trello.com/c/oViXS5Ty/893-credit-calculation-view-energy-densities-5

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Credit Calculation | View Energy Densities |5|](https://trello.com/c/oViXS5Ty/893-credit-calculation-view-energy-densities-5)

PR #1093, opened 2019-04-08
Energy densities and Carbon Intensities

amichard:
#1091 #1092 

Created the model, viewsets, serializers and tests

Changelog:
- Added Energy Density Model
- Added Carbon Intensity Model
- Added serializers
- Added viewsets
- Added tests


backend/api/models/EnergyDensity.py
```diff
@@ -0,0 +1,62 @@
...
+    )
+    density = models.DecimalField(
+        blank=True,
+        decimal_places=2,
+        default=Decimal('0.00'),
+        max_digits=5,
+        null=True,
+        db_comment="Energy density for the specific fuel class"
+    )
+    unit_of_measure = models.CharField(
```
klefler: @amichard, one thought... have you considered having the unit of measure values as a lookup value? I want to be cautious about recommending too much data model, but with these being repeatable values, it may be a good idea. Also, if these values need to come across into reports, or be used for calculations, it might help to have consistent values being used. Also, should this column be mandatory?

backend/api/models/CarbonIntensity.py
```diff
@@ -0,0 +1,57 @@
...
+        default=Decimal('0.00'),
+        max_digits=5,
+        null=True,
+        db_comment="Carbon Intensity (gCO2e/MJ) for the specific fuel class"
+    )
+
+    class Meta:
+        db_table = 'carbon_intensity'
+
+    db_table_comment = "Contains the carbon intensity for each approved " \
```
klefler: Is this for the "Default" CI? If so, would it make sense to add that to the table comment?

backend/api/models/EnergyDensity.py
amichard: I'll change this into a lookup table. Initially, I was debating about this. But it does make sense just to make it into a lookup table.

backend/api/models/CarbonIntensity.py
amichard: It seems like this is the default carbon intensity table. I'll add that in the table comment.

Issue #1098, opened 2019-04-11
Credit Calculation | Model Changes |5|


dainetrinidad:
There are some changes that we need to do for the models relating to the Energy Effectiveness Ratio, Energy Density and Default Carbon Intensity for the fuels.

Fuels might be grouped together based on the Regulation.
e.g. Biodiesel and HDRD both fall under Petroleum-based diesel fuel or renewable fuel in relation to diesel class fuel for Energy Effectiveness Ratio, but for energy density they're separated

Please refer to the checklist below on what we need to dohttps://trello.com/c/pFO1M5OL/1358-credit-calculation-model-changes-5

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Credit Calculation | Model Changes |5|](https://trello.com/c/pFO1M5OL/1358-credit-calculation-model-changes-5)

PR #1099, opened 2019-04-12
Credit calculation model changes

amichard:
#1098

Had to rearrange how the models are all associated. Some data in approved fuel types are supposed to be considered under a name based on the regulation when finding out what the energy density, energy effectiveness ratio, etc.

So I figured I'd just create additional models to categorized them

Changelog:
- Added Energy Density Category
- Added Energy Effectiveness Ratio Category
- Added Default Carbon Intensity Category
- Changed foreign key of energy density to the related category
- Changed foreign key of energy effectiveness ratio to the related category
- Changed foreign key of default carbon intensity to the related category
- Added new migrations
- Moved unit of measures into the Approved fuel type model
- Added migrations to populate the unit of measures
- Fixed the serializers, viewsets, tests and front-end


amichard:
[Credit Calculation | Model Changes |5|](https://trello.com/c/pFO1M5OL/1358-credit-calculation-model-changes-5)

Issue #1120, opened 2019-04-25
Fuel Codes | Indicate Part 2 renewable content % when adding or updating a fuel code |3|


amichard:
https://trello.com/c/daETtFu2/1333-fuel-codes-indicate-part-2-renewable-content-when-adding-or-updating-a-fuel-code-3#

**Story:**
As an IDIR user, I want to be able to specify the portion of the Part 2 fuel that is renewable versus non-renewable so that the system can correctly calculate the fuel's contribution to the renewable fuel requirement (part 2 requirement) when the fuel is reported in fuel supplier compliance reports. 

**Context and Intended Use:**
The Renewable Fuel Requirement (i.e. Part 2 requirement) requires fuel suppliers to supply a specified percentage of renewable fuel in relation to the gasoline and diesel fuel they supply. This is a volumetric, province-wide average requirement. For gasoline, fuel suppliers must ensure that there is 5% renewable content in the fuel, and for diesel, there must be 4% renewable content. 

Most fuels that have an approved fuel code are either 100% renewable (e.g. corn ethanol, soy biodiesel) or 0% renewable (e.g. fossil-based LNG). However, in a few instances, there are approved fuels that have a split percentage of renewable and non-renewable. In this case, only the renewable portion of the fuel counts towards the Part 2 requirement. 

The calculation of compliance with the Part 2 requirement is completed within the Compliance Report spreadsheet, and will be developed within the Compliance Reporting feature in TFRS. Therefore, for the time being, we need a way to specify the renewable content portion of each fuel code so that the system will know what portion of that fuel (0-100%) counts towards the Part 2 requirement within the Compliance Reporting feature.  

**Potential Design(s):**

- When adding or updating a fuel code, if an applicable fuel type is selected (see below), then have a check box that says "Is partially renewable?" appear to the right of the "Fuel" input field, with the default being unchecked (for new fuel codes). If this box is not checked, then assume 100% renewable.
- When this box is checked, a new input field appears (below it?) with the label "Percentage of Part 2 fuel that is renewable (%):". 

**Applicable Fuel types:**
This split renewable percentage feature only applies to certain fuel types:

- Ethanol
- Biodiesel
- HDRD
- Renewable gasoline 
- Renewable diesel
- Natural gas-based gasoline

**Data input & restrictions:**

- The data inputted into the field will be a percentage value, %
- Restrict values to between 0 to 100 (no negative values, max 100)
- Restrict value to two decimal places

**Forward looking usage:**

[...]

PR #1121, opened 2019-04-25
Renewable fuel codes

amichard:
#1120

Adds a checkbox and input field for renewable percentage

Changelog:
- Updated the Fuel code model to include renewable percentage
- Updated the Approved Fuel model to add an indicator on which fuel types can have partial 
- Adds a checkbox to the front-end to enable/disable renewable percentage (only applicable to certain fuel types)
- Updated the serializer to accommodate the new column


amichard:
[Fuel Codes | Indicate Part 2 renewable content % when adding or updating a fuel code |3|](https://trello.com/c/daETtFu2/1333-fuel-codes-indicate-part-2-renewable-content-when-adding-or-updating-a-fuel-code-3)

Issue #1125, opened 2019-04-30
Credit Calculation | Update Functionality should not get rid of old values |5|


dainetrinidad:
Updating a row for a credit calculation table shouldn't really "forget" the previous value, so instead of updating we really should be adding records instead.

e.g. CNG has a default carbon intensity of 64.65 with an effective for the year of 2017. If the values were supposed to change for 2018, then we want to make sure to keep the value for 2017. So we add a new record for 2018, instead. 

If the user has to make an edit for 2018, then we just add a new record again and when we have to get the value for 2018, we just get the most recent one and accept that as effective value.https://trello.com/c/ovjHPjn1/1382-credit-calculation-update-functionality-should-not-get-rid-of-old-values-5

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Credit Calculation | Update Functionality should not get rid of old values |5|](https://trello.com/c/ovjHPjn1/1382-credit-calculation-update-functionality-should-not-get-rid-of-old-values-5)

Issue #1124, opened 2019-04-30
Credit Calculation | Show current values in the credit calculation tables |5|


dainetrinidad:
**Story:**
As an IDIR user, I want to see the current values as they are today in the credit calculation tables

**Location:**

- IDIR user
- Administration tab, Compliance Reporting sub-tab

**Task:**
For the various identified tables, I want TFRS to show me the values that are effective today, even if there are new values that will become effective in the future. 

**Relevant Tables:**
- CI Limits
- Energy Effectiveness Ratio
- Default Carbon Intensities
- Energy Densities
- Fuel Classes https://trello.com/c/Z1mavdQe/1374-credit-calculation-show-current-values-in-the-credit-calculation-tables-5

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Credit Calculation | Show current values in the credit calculation tables |5|](https://trello.com/c/Z1mavdQe/1374-credit-calculation-show-current-values-in-the-credit-calculation-tables-5)

PR #1126, opened 2019-04-30
Credit calculation update changes

amichard:
#1124 #1125 

Changed how the update works, then refactored it as it was getting a lot of duplicate codes

Changelog:
- Created get in CreditCalculationService
- Created update in CreditCalculationService
- Updating a record, now creates a new one instead, and updates the expiry dates accordingly to avoid any overlaps or gaps
- Added tests to check if updates are working as expected


amichard:
[Credit Calculation | Update Functionality should not get rid of old values |5|](https://trello.com/c/ovjHPjn1/1382-credit-calculation-update-functionality-should-not-get-rid-of-old-values-5)

amichard:
[Credit Calculation | Show current values in the credit calculation tables |5|](https://trello.com/c/Z1mavdQe/1374-credit-calculation-show-current-values-in-the-credit-calculation-tables-5)

Issue #1133, opened 2019-05-02
Credit Transactions | React Table Optimization Part 2 Remember selected organization |2|


amichard:
https://trello.com/c/hXefjlKE/1369-credit-transactions-react-table-optimization-part-2-remember-selected-organization-2#

**Story:**
As an IDIR user, I want the Credit Transactions react table to remember the organization I have selected when using the back button after viewing a credit transaction so that I can easily view organization-specific transactions within the system. 

**Scenario:**
GIVEN that I am an IDIR user
AND I am on the Credit Transactions page
AND I have selected an organization from the drop-down menu (i.e. "show transactions involving:") 
AND I am viewing the details of a particular credit transaction entry (i.e. transaction view page)
WHEN I click on the back button
THEN I am returned to the main credit transactions page 
AND the organization is still selected within the drop-down menu (i.e. the table is still filtering transactions based on the selected organization

**Details:**
Currently when using the back button, the credit transactions react table remembers a user's position within the table (i.e. which page you are on) and any filtering that was used within the react table (e.g. transaction type = Part 3 Award). However, for IDIR users, the table does not remember if an organization was selected from the drop down menu in the top right hand portion of the page (e.g. "Show transactions involving: TFRS Fantastic Fuels"). This selection resets every time the user uses the back button when viewing a specific transaction. This card is to have TFRS remember the selected organization so that the user does not have to continually re-select them every time they use the back button.

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Credit Transactions | React Table Optimization Part 2 Remember selected organization |2|](https://trello.com/c/hXefjlKE/1369-credit-transactions-react-table-optimization-part-2-remember-selected-organization-2)

PR #1134, opened 2019-05-02
Organization ID is now being saved when filtering in credit transactions page

amichard:
#1133 

I piggy backed on the credit-transfers state, as it's sharing the variables anyway.

Changelog:
- Automatically save the organization in the credit-transfers state when selecting an organization
- Auto select organization in drop-down in credit transactions
- Fixed a potential issue in the credit transaction history when something is declined and the record didn't have any roles attached to it (old records)
- ESLint fixes


amichard:
[Credit Transactions | React Table Optimization Part 2 Remember selected organization |2|](https://trello.com/c/hXefjlKE/1369-credit-transactions-react-table-optimization-part-2-remember-selected-organization-2)

Issue #1152, opened 2019-05-13
Compliance Report C | An additional column appears when Other is selected for Expected Use |2|


dainetrinidad:
**Story:**
As a BCeID user, I want the system to prompt me to provide the details about Fuels used for other purposes when I select the Other category so that I can ensure that I am fulfilling my reporting obligation. 

**Scenario:**
GIVEN that I am a BCeID user
AND I have the Compliance Reporting role [Card 1299](https://trello.com/c/X6pY9KoM/1299-compliance-report-create-compliance-reporting-role)
AND I am filling out Schedule C of the Compliance Report (new report or editing a draft)
AND I am filling out a particular entry row within Schedule C
AND I have clicked on the Expected Use cell
AND I can see the list of predefined options [Card 1301](https://trello.com/c/cCaMiWSZ/1301-compliance-report-c-provide-predefined-selections-for-expected-use-column-0)
WHEN I select the Other option
THEN a new input field/cell/column appears to the right of the Expected Use column
AND the system prompts me to "Enter in the expected use"
AND I am able to enter an explanation of the "Other" fuel use within that input field / cell / column

**Potential Design(s):**
Similar to the zero dollar option menu that appears when a user enters a zero credit price (affiliate or other), when the user selects the Other category have an additional entry field appear to the right of the expected use cell that allows the user to enter in the details of the "Other" fuel use. 

**Additional Requirements:**

- This input field should be free text (allow anything)
- The user should be required to enter something into this cell if they have selected the Other category in the Expected Use column

https://trello.com/c/eUq1jomb/1302-compliance-report-c-an-additional-column-appears-when-other-is-selected-for-expected-use-2

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Compliance Report C | An additional column appears when Other is selected for Expected Use |2|](https://trello.com/c/eUq1jomb/1302-compliance-report-c-an-additional-column-appears-when-other-is-selected-for-expected-use-2)

Issue #1151, opened 2019-05-13
Compliance Report C | Provide predefined selections for Expected Use column |3|


dainetrinidad:
**Story:**
As a BCeID user, I want the system to provide me with suggestions for the expected use of Fuels used for other purposes so that I can ensure that I am reporting accurately. 

**Scenario:**
GIVEN that I am a BCeID user
AND I have the Compliance Reporting role [Card 1299](https://trello.com/c/X6pY9KoM/1299-compliance-report-create-compliance-reporting-role)
AND I am filling out Schedule C of the Compliance Report (new report or editing a draft)
AND I am filling out a particular entry row within Schedule C
WHEN I click on the Expected Use cell
THEN a list of predefined options appear
AND I am able to select one of the options

**Options:**
The following is a list of available options for the Expected Use column:

- Heating oil
- Department of National Defence (Canada)
- Aviation
- Otherhttps://trello.com/c/cCaMiWSZ/1301-compliance-report-c-provide-predefined-selections-for-expected-use-column-3

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Compliance Report C | Provide predefined selections for Expected Use column |3|](https://trello.com/c/cCaMiWSZ/1301-compliance-report-c-provide-predefined-selections-for-expected-use-column-3)

PR #1153, opened 2019-05-13
Schedule C - Expected Use and Other column

amichard:
#1151 #1152 #1154

Added options for Expected Use and conditions for Other column

Changelog:
- Added model and migration for Expected Uses
- Added viewset, serializer for Expected Uses
- Loading options for Expected Use in the spreadsheet
- Added a trigger to enable/disable the last column as needed
- Added row number
- Changed hard-coded column number to use constants instead
- Added totals as a floating object on the lower right portion of the page
- Allow adding multiple rows


backend/api/models/ExpectedUse.py
```diff
@@ -0,0 +1,47 @@
...
+        unique=True,
+        db_comment="Description on what the fuel is going to be used for."
+    )
+
+    objects = ExpectedUseManager()
+
+    class Meta:
+        db_table = 'expected_use'
+
+    db_table_comment = "Different options for the Expected Use column in " \
```
klefler: Is it possible to expand the description? It's not clear what "expected use" is... and maybe clarify the name of the "Schedule C" report? Or may not even need to specify in the description the specific use for that schedule... if there's a chance this code table could be used for other things.

backend/api/models/ExpectedUse.py
```diff
@@ -0,0 +1,47 @@
...
+    description = models.CharField(
+        max_length=100,
+        unique=True,
+        db_comment="Description on what the fuel is going to be used for."
+    )
+
+    objects = ExpectedUseManager()
+
+    class Meta:
+        db_table = 'expected_use'
```
klefler: If this is a lookup table, should the table name be similar to others? i.e. "expected_use_type"

amichard: sure i can name it to expected_use_type

backend/api/models/ExpectedUse.py
amichard: ill try to provide a better explanation on what the table does

amichard:
[Compliance Report C | Show the total fuel excluded for each fuel class |2|](https://trello.com/c/OdveyJjN/1303-compliance-report-c-show-the-total-fuel-excluded-for-each-fuel-class-2)

amichard:
[Compliance Report C | An additional column appears when Other is selected for Expected Use |2|](https://trello.com/c/eUq1jomb/1302-compliance-report-c-an-additional-column-appears-when-other-is-selected-for-expected-use-2)

amichard:
[Compliance Report C | Provide predefined selections for Expected Use column |3|](https://trello.com/c/cCaMiWSZ/1301-compliance-report-c-provide-predefined-selections-for-expected-use-column-3)

PR #1155, opened 2019-05-14
Add Autosave Support

plasticviking:
Add an HOC, `autosaved` which wraps other components to give them the ability to persist autosaved data (`state`, most likely) on the server side.

Takes a config object as a parameter, with the form
`{ name, key, version }`, and provides wrapped classes with four new props:
name should be the type of form.
version should change if there are breaking changes to the form
key should represent the unique instance being edited (eg '1' or 'unsaved')

`updateStateToSave` is a function that takes a state that should be persisted (call it in `handleInputChange` or similar)
`updateAutosaveKey` is a function taking a key and will update `key` if necessary
`loadedState` is the state retrieved from the backend, if any
`saving` is true when a save is in progress (for displaying and icon)

Loads before rendering the wrapped component and passes `loadedState` as an initial prop
Saves on `componentWillUnmount`. It would be trivial to have it save every 30s or something too.

Added backend support including expiry of old entries

Wired up `FuelCodeAddContainer` as a demo. Try creating a new fuel code to see it work.


amichard:
Looks good!

I am finding that it's not autosaving if I close the browser instead of navigating away.

kuanfandevops:
@plasticviking 0129 migration has been existed, so the two migrations in this PR need to be renamed.

kuanfandevops:
I'll get it merged first and rename the migration later on

PR #1160, opened 2019-05-15
Development environment improvements

plasticviking:
#Changelog:
- Replace mailslurper with a basic python SMTP logger
- Add more compliance periods to `dockerized` ops scripts and bundle the credit calculation scripts into it
- Fix `operational` script `0008` to deal with renamed energy density categories
- Add `react-hot-loader` dependency and integrate with docker webpack config
- Chunk splitting and HTML generation for dev-server to improve compile times
- Minio bucket created on startup if it doesn't exist

### DIFF
--- a/backend/api/app.py
+++ b/backend/api/app.py
@@ -30,7 +30,7 @@ from pika.exceptions import AMQPError
 from django.apps import AppConfig
 from django.db.models.signals import post_migrate
 from minio import Minio
-from minio.error import MinioError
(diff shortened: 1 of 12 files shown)


kuanfandevops:
@plasticviking the operational script 0008_add_default_carbon_intensities.py is updated, do we need to run it? the latest script in 0012..

kuanfandevops:
message from developers, no need to run the updated script

Issue #1162, opened 2019-05-16
User Management | Adding fuel supplier user shows all roles including government roles |1|


dainetrinidad:
- IDIR user, Admin role
- Fuel Suppliers nav link, then choose specific fuel supplier org, then click on new user 

All roles are shown, including government roles; since this is a fuel supplier user, we should only show fuel supplier roleshttps://trello.com/c/OQjd2mVN/1436-user-management-adding-fuel-supplier-user-shows-all-roles-including-government-roles-1

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [User Management | Adding fuel supplier user shows all roles including government roles |1|](https://trello.com/c/OQjd2mVN/1436-user-management-adding-fuel-supplier-user-shows-all-roles-including-government-roles-1)

PR #1163, opened 2019-05-16
Government roles should only show up when creating IDIR users

amichard:
#1162

I think we accidentally removed the auto-filter. This should add back the logic to show the roles based on the user being created.


amichard:
[User Management | Adding fuel supplier user shows all roles including government roles |1|](https://trello.com/c/OQjd2mVN/1436-user-management-adding-fuel-supplier-user-shows-all-roles-including-government-roles-1)

PR #1226, opened 2019-06-17
Frontend Refactoring

plasticviking:
# Changelog
 - Create a `ComplianceReportingEditContainer` container to hold all schedules and push state up to this level
 - Schedule A saving now works as expected, and it should now be much simpler to implement the remaining schedules
 - Child components receive `complianceReport` in a prop

### DIFF
--- a/frontend/src/router.js
+++ b/frontend/src/router.js
@@ -52,11 +52,6 @@ import UserAddContainer from './admin/users/UserAddContainer';
 import UserEditContainer from './admin/users/UserEditContainer';
 import NotFound from './app/components/NotFound';
 import ComplianceReportingContainer from './compliance_reporting/ComplianceReportingContainer';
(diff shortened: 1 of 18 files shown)


frontend/src/router.js
```diff
@@ -329,56 +325,22 @@ const Router = routerProps => (
...
-          <Route
-            key="compliance_reporting_edit_schedule_c"
-            path={COMPLIANCE_REPORTING.EDIT_SCHEDULE_C}
-            component={withRouter(withReferenceData()(ScheduleCContainer))}
-          />,
-          <Route
-            key="compliance_reporting_edit_schedule_d"
-            path={COMPLIANCE_REPORTING.EDIT_SCHEDULE_D}
-            component={withRouter(withReferenceData()(ScheduleDContainer))}
+            exact={false}
```
amichard: Shouldn't this default to false?

Issue #1235, opened 2019-06-20
Login | Login Page not displaying on some circumstances |1|


dainetrinidad:
Matthew encountered an issue with the login page where it wasn't displaying either of the buttons. 

Upon inspection I noticed that there's a console error and that might be affecting it. https://trello.com/c/2KNfZgW5/1402-login-login-page-not-displaying-on-some-circumstances-1

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Login | Login Page not displaying on some circumstances |1|](https://trello.com/c/2KNfZgW5/1402-login-login-page-not-displaying-on-some-circumstances-1)

KMenke:
Resolved in Sprint Quasar.

PR #1238, opened 2019-06-20
Renamed Login buttons classes

amichard:
#1235 

This is so that ad blockers won't hide the login buttons.

Also, removing release.md as it's not being used.


amichard:
[Login | Login Page not displaying on some circumstances |1|](https://trello.com/c/2KNfZgW5/1402-login-login-page-not-displaying-on-some-circumstances-1)

PR #1273, opened 2019-06-27
Schedule D Backend

plasticviking:
# Changelog
 - Added Schedule D models and serializers


kuanfandevops:
@plasticviking looks like there is a conflict in this pr

backend/api/models/ComplianceReportSchedules.py
```diff
@@ -151,3 +154,123 @@ class ScheduleBRecord(Model):
...
+
+    class Meta:
+        db_table = 'compliance_report_schedule_d_sheet_input'
+
+
+class ScheduleDSheetOutput(Model):
+
+    class OutputCells(Enum):
+        """
+        Enum of possible output cell names
```
klefler: Should this be a lookup table? I'd prefer to see a lookup table rather than enums, but happy to discuss... Is this a case where a separate table doesn't make sense?

klefler:
Hi Robert. Question regarding commenting of the tables and columns... is it possible to please add the table/column descriptions for these new compliance report tables? This helps to keep the rest of the model consistent, and provides an understanding of what these tables/columns are storing.

backend/api/models/ComplianceReportSchedules.py
```diff
@@ -151,3 +154,123 @@ class ScheduleBRecord(Model):
...
+        max_length=100,
+        blank=True,
+        null=True
+    )
+    value = models.CharField(
+        max_length=100,
+        blank=True,
+        null=True
+    )
+    units = models.CharField(
```
klefler: Would there be any benefit to connecting to the unit_of_measure table for these values? It would help keep the data consistent. Might be low risk, but just wanted to mention it.

backend/api/models/ComplianceReportSchedules.py
plasticviking: Hi Kristin,

Generally I would say that I'd prefer to use lookup tables, too. We had a discussion about this functionality and decided that any future changes would require a code update which is the only reason I hard-coded them. The integrity can be validated in-app rather than making a round-trip to the DB. That said, I'm happy to change it to a lookup table if you prefer.

I think we have a larger technical debt/architecture problem in that we use lookup tables inappropriately in some cases. For example, the status of credit transfers should, in my opinion, be an application Enum and not a table since application logic changes significantly for a credit transfer depending on the status it's in.

This is not one of those cases, though -- I just thought these data would change infrequently enough to not warrant a separate table.

plasticviking:
Hi @klefler. I have added table and column comments in the latest push for this

backend/api/models/ComplianceReportSchedules.py
klefler: Thanks @plasticviking, Yes, I think I'd prefer that we look at options for making this example a look up table, and I'm definitely happy to discuss those examples where you feel an application enum is a better design. In NRM applications, enums are not typically used due to a number of reasons, and if they are, I believe it is a rare case. A decision to go with a different pattern could definitely be considered if there is risk in having the design as-is. We can chat when you're back! Thanks.

PR #1277, opened 2019-06-28
Update Openshift templates

kuanfandevops:
Templates updated in order to smoothly restore the environment
Split the pipeline scripts into various groovy file


kuanfandevops:
close without merging

PR #1284, opened 2019-07-02
Schedule buttons

amichard:
#1285 

Changelog:
-Schedule D tab names is now based on Fuel Type and Carbon Intensity
-Moved the buttons into the main container so there's less repeating code.
-Removed the unnecessary required props after.
-ESLint fixes


amichard:
[Compliance Report D | Minor adjustments to sub tabs when there is more then one fuel being reported |1|](https://trello.com/c/oXX8pMv9/1516-compliance-report-d-minor-adjustments-to-sub-tabs-when-there-is-more-then-one-fuel-being-reported-1)

PR #1314, opened 2019-07-17
Reverse migration that allows Schedule D to be saved with null fuel type and fuel class

amichard:
Reversing the migration and fuel type and fuel class change.

After talking to Rob, I now understand why the field was not nullable in the first place.


plasticviking:
I'm not sure I like this -- In my opinion, autosave is the place for not-yet-valid data. I think trying to save a record without these values _**should**_ result in a validation failure.

Issue #1327, opened 2019-07-22
Compliance Report B | Some active fuel codes are not selectable within Schedule B |2|


amichard:
**Issue:**
Some active fuel codes are not showing up in the fuel code drop-down menu within Schedule B despite being active during the compliance period in which the report relates. "Active" in this case means that the fuel code has an effective period that overlaps with the compliance period in which the compliance report relates. 

**Example:**

- Dev environment
- Fuel code BCLCF133.2; HDRD fuel type, effective period: February 1, 2017 to December 31, 2020
- When reporting the supply of fuel within **Schedule B** of the **Compliance Report for 2018**, after the user selects HDRD and fuel code as the Provision of the Act, the fuel code BCLCF133.2 does not show up as selectable in the fuel code list. 

https://trello.com/c/5ujzXy9I/1618-compliance-report-b-some-active-fuel-codes-are-not-selectable-within-schedule-b-2

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Compliance Report B | Some active fuel codes are not selectable within Schedule B |2|](https://trello.com/c/5ujzXy9I/1618-compliance-report-b-some-active-fuel-codes-are-not-selectable-within-schedule-b-2)

KMenke:
Resolved in Sprint Supermassive

PR #1328, opened 2019-07-22
Fixed Missing Fuel Codes

amichard:
#1327 

Changelog:
- Added another condition to take into account if the fuel code effective date and expiry date is greater than the compliance period dates


amichard:
[Compliance Report B | Some active fuel codes are not selectable within Schedule B |2|](https://trello.com/c/5ujzXy9I/1618-compliance-report-b-some-active-fuel-codes-are-not-selectable-within-schedule-b-2)

Issue #1329, opened 2019-07-22
Compliance Report B | Save function does not save user inputted CI when relying on Alternative Method provision |5|


amichard:
**Issue:**
When the user relies on the Alternative Method provision and has inputted a CI into the CI of Fuel cell, the inputted CI value does not save when the user uses the save feature (save button). Instead, the value is cleared and the cell becomes uneditable. 

**Desired state:**
When the user saves, the inputted CI of Fuel value is saved and the cell remains editable in case the user wants to adjust the inputted CI value. 

https://trello.com/c/8xIjfLVU/1619-compliance-report-b-save-function-does-not-save-user-inputted-ci-when-relying-on-alternative-method-provision-5

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Compliance Report B | Save function does not save user inputted CI when relying on Alternative Method provision |5|](https://trello.com/c/8xIjfLVU/1619-compliance-report-b-save-function-does-not-save-user-inputted-ci-when-relying-on-alternative-method-provision-5)

KMenke:
Resolved in Sprint Supermassive

PR #1330, opened 2019-07-22
Alternative carbon intensity

amichard:
#1329 

Rob already fixed the issue with saving and loading the carbon intensity when alternate method is selected. This is so that the page loads it as an input field and fixes the summary page.

Changelog:
- Should now load carbon intensity as an input field when alternate method is selected
- Summary screen now takes into account GHGenius and Alternate methods
- Fixed a caching issue with GHGenius options when creating a new compliance report

### DIFF
--- a/frontend/src/compliance_reporting/ComplianceReportingEditContainer.js
+++ b/frontend/src/compliance_reporting/ComplianceReportingEditContainer.js
@@ -185,6 +185,8 @@ class ComplianceReportingEditContainer extends Component {
       return (<Loading />);
     }
 
+    let item = {};
+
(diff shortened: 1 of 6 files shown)


amichard:
[Compliance Report B | Save function does not save user inputted CI when relying on Alternative Method provision |5|](https://trello.com/c/8xIjfLVU/1619-compliance-report-b-save-function-does-not-save-user-inputted-ci-when-relying-on-alternative-method-provision-5)

amichard:
No problem, Rob!

amichard:
Rob has fixed these issues with his PR

Issue #1426, opened 2019-08-29
Compliance Report | Optimize save submit when there are validation errors |1|


amichard:
**Task:**

This card is to "optimize" the save and submit functions when there validation errors. 

**Save:**

- Current state: button is greyed out and tooltip appears with back-end language
- Desired state: button is greyed out, tooltip says "Please fix validation errors before saving."

**Submit:**

- Current state: button is clickable, receive "Error saving" system alert after completing declaration and trying to submit
- Desired state: button is greyed out, tooltip says "Please fix validation errors before submitting."

https://trello.com/c/XoKdxQja/1965-compliance-report-optimize-save-submit-when-there-are-validation-errors-1

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Compliance Report | Optimize save submit when there are validation errors |1|](https://trello.com/c/XoKdxQja/1965-compliance-report-optimize-save-submit-when-there-are-validation-errors-1)

KMenke:
resolved in sprint vulcan

PR #1427, opened 2019-08-29
Save and Submit button changes

amichard:
#1426 

Changelog:
- Changed the validation message for the save button
- Submit button validates similarly to the save button


amichard:
[Compliance Report | Optimize save submit when there are validation errors |1|](https://trello.com/c/XoKdxQja/1965-compliance-report-optimize-save-submit-when-there-are-validation-errors-1)

PR #1472, opened 2019-09-10
Compliance Reporting Containers prefer snapshot values

plasticviking:
Compliance Reporting containers will refer to snapshot rather than computed values when these values exist.

Trello card 1954

### DIFF
--- a/frontend/src/compliance_reporting/ScheduleBContainer.js
+++ b/frontend/src/compliance_reporting/ScheduleBContainer.js
@@ -3,8 +3,8 @@
  * All data handling & manipulation should be handled here.
  */
 
-import React, { Component } from 'react';
-import { connect } from 'react-redux';
+import React, {Component} from 'react';
+import {connect} from 'react-redux';
 import PropTypes from 'prop-types';
 import FontAwesomeIcon from '@fortawesome/react-fontawesome';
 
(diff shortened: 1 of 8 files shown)


frontend/src/compliance_reporting/ScheduleCContainer.js
```diff
@@ -124,65 +124,85 @@ class ScheduleCContainer extends Component {
...
+  componentWillReceiveProps(nextProps, nextContext) {
+    const {grid} = this.state;
 
-    if (nextProps.scheduleState.scheduleC && nextProps.scheduleState.scheduleC.records) {
-      if ((grid.length - 2) < nextProps.scheduleState.scheduleC.records.length) {
-        this._addRow(nextProps.scheduleState.scheduleC.records.length - (grid.length - 2));
+    if (nextProps.snapshot) {
+      const source = nextProps.snapshot.scheduleC;
+
+      if ((grid.length - 2) < source.records.length) {
```
amichard: Can we add a check for source here? and to make sure we can retrieve length from source.records?

I tested this with a compliance report that doesn't have schedule B filled in and it's running into errors

frontend/src/compliance_reporting/ScheduleBContainer.js
```diff
@@ -167,77 +167,112 @@ class ScheduleBContainer extends Component {
...
-      return;
-    }
+    if (nextProps.snapshot) {
+      // just use the snapshot
+      const source = nextProps.snapshot.scheduleB;
 
-    if ((grid.length - 2) < nextProps.scheduleState.scheduleB.records.length) {
-      this._addRow(nextProps.scheduleState.scheduleB.records.length - (grid.length - 2));
-    }
+      if ((grid.length - 2) < source.records.length) {
```
amichard: Can we add a check for source here? and to make sure we can retrieve length from source.records?

I tested this with a compliance report that doesn't have schedule B filled in and it's running into errors

plasticviking:
@amichard I've added those checks

PR #1485, opened 2019-09-17
Create supplemental report

plasticviking:
# Changelog
- Create a supplemental report button and action
- Copies all schedule and summaries (but not snapshot or status) into new report
- Also added `nickname` field (but the serializers don't write to it yet)
- Modify list view to show tree structure for supplemental reports
- Tests

## Pending in next PR:
- Probably some style changes (I'll await feedback)
-Changelog for supplemental reports
-More obvious visual feedback when you're working on a report which has supplemental reports (or is itself a supplemental report)
- History records need to be adjusted
- Edit the nickname


backend/api/serializers/ComplianceReport.py
```diff
@@ -579,9 +626,106 @@ def validate_status(self, value):
     def create(self, validated_data):
         status_data = validated_data.pop('status')
         status = ComplianceReportWorkflowState.objects.create(**status_data)
-        return ComplianceReport.objects.create(
+        cr = ComplianceReport.objects.create(
             status=status,
             **validated_data)
+        print('c')
```
amichard: debug code?

amichard:
Looks good other than the debug codes

PR #1513, opened 2019-09-27
Sort Order for compliance reports

plasticviking:
# Changelog
- Default sort order for compliance reports is "compliance period, last activity" where last activity is the latest update date for any report in the chain of supplementals

### DIFF
--- a/frontend/src/compliance_reporting/components/ComplianceReportingTable.js
+++ b/frontend/src/compliance_reporting/components/ComplianceReportingTable.js
@@ -99,13 +99,22 @@ class ComplianceReportingTable extends Component {
       Header: 'Status',
       id: 'status',
       minWidth: 75
-    }, {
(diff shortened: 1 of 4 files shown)


frontend/src/compliance_reporting/components/ComplianceReportingTable.js
```diff
@@ -146,7 +155,8 @@ class ComplianceReportingTable extends Component {
         subRowsKey={'supplementalReports'}
 
         getTrProps={(state, row) => {
-          const stripeClass = row && row.nestingPath[0] % 2 ? 'odd':'even' || 'even';
+          console.log(row ? row.nestingPath : '');
```
amichard: Looks like a debug code?

plasticviking: Fixed

PR #1518, opened 2019-10-01
Fix unit tests

plasticviking:
# Changelog
- Correct an issue with unit test fixtures for compliance reports
- Add a new `docker-compose-test.yml` file as an example of how to run unit tests in a container
- Remove dependency on Minio for tests


kuanfandevops:
@plasticviking 
looks like minio is not in docker-compose-test.yml, is the volume minio_data still needed?

PR #1522, opened 2019-10-03
Supplemental view changes

plasticviking:
# Changelog
- Add submissions history tab for compliance reports
- Clicking on any supplemental report brings you to the latest one


amichard:
Some stuff I noticed:
- Collapsible history is now static with bullet points. Was this intended?
- Snapshot button is removed. Is Review and history supposed to replace it?
- Deleted supplemental reports are being shown and clicking on the report causes the page to throw an error

PR #1527, opened 2019-10-04
Supplemental require comment

plasticviking:
# Changelog
- Supplemental report submissions will require a 1-500 character
explanation
 - Fixed a unit test defect that arose as part of a previous PR
 - Fixed a frontend defect where readonly reports were calling
validation unnecessarily.


amichard:
Looks good. Will mark as approved once #1522 gets approved

PR #1528, opened 2019-10-04
Compliance report list updates

plasticviking:
# Changelog
 - Update compliance report list table to match desired specs (include supplemental status column, drop tree view)


amichard:
Waiting for #1522 to be merged so I can double-check some stuff here

Issue #1529, opened 2019-10-07
Compliance Report B | Restrict the use of Lines 6 8 17 19 |2|


amichard:
**Task:**
This card is to implement additional restrictions on Lines 6, 8, 17, 19 of the Summary page of the compliance report. 

**Note:** some restrictions have already been implemented for Line 6 and 17, and for Line 8 and 19:

- **Line 6 and 17**: [Card 1474](https://trello.com/c/MQc801ue/1474-compliance-report-part-2-summary-line-6-and-line-17-2-v1414) restricted the value that can be entered into this field to be no more than 5% of Line 4/15
- **Line 8 and 19**: [Card 1478](https://trello.com/c/M7ErZxkL/1478-compliance-report-part-2-summary-line-8-and-line-19-2-v1414) restricted the value that can be entered into this field to be no more than 5% of Line 4/15

**Line 6:**

- The cell should only be **editable** when Line 2 > Line 4. If Line 2 < Line 4, then the cell should **not be editable**.
- If Line 2 > Line 4, the value entered should be restricted to no more than the **lower** of the two values:
 - Line 2 - Line 4
 - 5% of Line 4

**Line 17:**

- The cell should only be **editable** when Line 13 > Line 15. If Line 13 < Line 15, then the cell should **not be editable**.
- If Line 13 > Line 15, the value entered should be restricted to no more than the **lower** of the two values:
 - Line 13 - Line 15
 - 5% of Line 15

**Line 8:**

- The cell should only be **editable** when Line 4 > Line 2. If Line 4 < Line 2, then the cell should **not be editable**.
- If Line 4 > Line 2, the value entered should be restricted to no more than the **lower** of the two values:
 - Line 4 - Line 2
 - 5% of Line 4

**Line 19:**

- The cell should only be **editable** when Line 15 > Line 13. If Line 15 < Line 13, then the cell should **not be editable**.
- If Line 15 > Line 13, the value entered should be restricted to no more than the **lower** of the two values:
 - Line 15 - Line 13
 - 5% of Line 15

**Additional Requirements:**

- When one or more of these cells switch from an editable cell to a non-editable cell, the values (if any) within the non-editable cell should be cleared. This will allow us to properly handle a situation where the user enters a value into one or more of these cells, then goes back to Schedule B and changes some information which causes one or more of these cells to switch from editable to non-editable
- All four of these cells should not be editable if there were no Part 2 fuel supplied by the supplier, as reported in Schedule B and Schedule C. 
[...]

PR #1530, opened 2019-10-07
Schedule Summary Restrictions

amichard:
#1529 

Changelog:
- Only 1 field is enabled per section in the Summary Page
- Values get automatically adjusted if the max value is lower than the inputted value
- Fixed some bugs with how the state is managed, so there should be less issues when switching back and forth to the summary page


amichard:
[Compliance Report B | Restrict the use of Lines 6 8 17 19 |2|](https://trello.com/c/A0rIN4zz/1727-compliance-report-b-restrict-the-use-of-lines-6-8-17-19-2)

amichard:
[Bug | Squash em |2|](https://trello.com/c/fYK7rT5E/2046-bug-squash-em-2)

PR #1532, opened 2019-10-08
Open unit test cases after database is switched to Postgresql from default SQLite

kuanfandevops:


kuanfandevops:
close it as the release tag need to be changed

Issue #1533, opened 2019-10-08
Compliance Report | Can only Submit from Summary page |2|


amichard:
Fuel supplier user can only submit from the Summary page. 

- Also add validation panel to the summary page

https://trello.com/c/0z5egRhm/2045-compliance-report-can-only-submit-from-summary-page-2

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Compliance Report | Can only Submit from Summary page |2|](https://trello.com/c/0z5egRhm/2045-compliance-report-can-only-submit-from-summary-page-2)

KMenke:
resolved

PR #1534, opened 2019-10-08
Submit can only be clicked in the summary page

amichard:
#1533 

Changelog:
- Submit button is disabled until you get to the Summary page
- Added validation messages to the Summary page
- Computation in Summary is disabled if there are validation issues


amichard:
[Compliance Report | Can only Submit from Summary page |2|](https://trello.com/c/0z5egRhm/2045-compliance-report-can-only-submit-from-summary-page-2)

PR #1535, opened 2019-10-08
Update to enable more unit test cases

kuanfandevops:


kuanfandevops:
@plasticviking  I created to PR and assigned you to review
Don't worry the following lines as they are for unit test
* envVar(key: 'DATABASE_USER', value: 'tfrstest'),
* envVar(key: 'DATABASE_PASSWORD', value: 'tfrstest')
Also tfrsRelease="devops-psql-for-unittest" is ok, I'll update it when deploy v1.5.9

PR #1540, opened 2019-10-09
Fixed summary page and exclusion report

amichard:
Changelog:
- Fixed issue with summary page
- Temporary fix for broken exclusion report


frontend/src/compliance_reporting/ScheduleSummaryContainer.js
```diff
@@ -397,14 +388,19 @@ class ScheduleSummaryContainer extends Component {
...
-      this.setState({
-        diesel,
-        gasoline,
-        part3,
-        penalty
-      });
     }
+
+    console.error(diesel);
+    console.error(gasoline);
```
plasticviking: intentional?

amichard: thanks for catching this. should be gone now

PR #1565, opened 2019-10-23
Exclusion report supplemental reporting

plasticviking:
# Changelog
- Added exclusion report supplemental reporting including history and changelogs

Trello cards:
1743, 1991, 1760, 1999, 1755, 1993


backend/api/serializers/ComplianceReport.py
```diff
@@ -933,6 +934,22 @@ def create(self, validated_data):
                 summary.credits_offset = original_summary.credits_offset
                 summary.save()
 
+            if original_report.exclusion_agreement is not None:
+                exclusion_agreement = ExclusionAgreement.objects.create()
+                cr.exclusion_agreement = exclusion_agreement
+                exclusion_agreement.save()
+                exclusion_agreement.save()
```
amichard: this looks duplicated

plasticviking: Good catch. fixed.

Issue #1581, opened 2019-10-29
Credit Calculation | Show CI Limit values in the UI for 2013 14 2015 and 2016 |1|


amichard:
https://trello.com/c/pjBBzCi4/2037-credit-calculation-show-ci-limit-values-in-the-ui-for-2013-14-2015-and-2016-1

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Credit Calculation | Show CI Limit values in the UI for 2013 14 2015 and 2016 |1|](https://trello.com/c/pjBBzCi4/2037-credit-calculation-show-ci-limit-values-in-the-ui-for-2013-14-2015-and-2016-1)

PR #1582, opened 2019-10-29
CI Limits for 2013-2016

amichard:
#1581 

Changelog:
- New operational script to update the expiration dates for 2013-2016 so it shows up in the table


amichard:
[Credit Calculation | Show CI Limit values in the UI for 2013 14 2015 and 2016 |1|](https://trello.com/c/pjBBzCi4/2037-credit-calculation-show-ci-limit-values-in-the-ui-for-2013-14-2015-and-2016-1)

Issue #1592, opened 2019-11-04
Compliance Report | Copy Paste should not be case sensitive |2|


amichard:
**Task:**
This card is to have the compliance reporting spreadsheet plug-in not be case sensitive when copying and pasting values. For example, biodiesel should be accepted as a fuel type as well as Biodiesel. Currently only the latter is accepted when copying and pasting. 

**Affected Schedules & Columns:**
This applies to the following columns:

- Schedule B: Fuel Type, Fuel Class, Provision of the Act
- Schedule C: Fuel Type, Fuel Class

https://trello.com/c/m1o7sLQE/1647-compliance-report-copy-paste-should-not-be-case-sensitive-2

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Compliance Report | Copy Paste should not be case sensitive |2|](https://trello.com/c/m1o7sLQE/1647-compliance-report-copy-paste-should-not-be-case-sensitive-2)

PR #1593, opened 2019-11-04
Compliance reporting copy paste

amichard:
#1592 

Changelog:
- Copying and pasting is no longer case sensitive
- Also, fixed an issue where you can paste in any value into the cells that have dropdowns


amichard:
[Compliance Report | Copy Paste should not be case sensitive |2|](https://trello.com/c/m1o7sLQE/1647-compliance-report-copy-paste-should-not-be-case-sensitive-2)

PR #1601, opened 2019-11-06
Changed the font so it uses BC Sans

amichard:
#1602 

Changelog:
- Changed the font family to prioritize BC Sans
- Not happy about the !important, but there are too many things that are overriding the font family


amichard:
[Font | Change our font type to BC Sans Typeface |2|](https://trello.com/c/7m1p50vU/2082-font-change-our-font-type-to-bc-sans-typeface-2)

PR #1614, opened 2019-11-14
Change Archive to Delete in Notifications

amichard:
Changelog:
- Renamed archive to delete
- Changed the icon so it's the same as the other delete icons
- Changed the delete icon in historical data entry
- Also, coloured it red


amichard:
[Notifications | Change Archive label to Delete |0|](https://trello.com/c/6S1plnnQ/2096-notifications-change-archive-label-to-delete-0)

Issue #1618, opened 2019-11-15
Reports | Prevent the selection of one's own organization when completing compliance and exclusion reports |1|


amichard:
**Issue:**
Currently, a fuel supplier is able to select themselves from the list of recognized fuel suppliers when filling out the different portions of a compliance and exclusion report. For example, a fuel supplier can select themselves as their transaction partner in Schedule A; we want to prevent this and these types of reported transactions cannot occur between one single organization. 

**Restriction measure:**

- Throw a validation error? 
- Remove one's own organization from being an option in the autosuggest feature

**Affected areas:**

- Compliance Report, Schedule A: Legal Name of Trading Partner column
- Exclusion Report, Exclusion Agreement tab: Legal Name of Transaction Partner column

https://trello.com/c/sslrUM28/2093-reports-prevent-the-selection-of-ones-own-organization-when-completing-compliance-and-exclusion-reports-0

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Reports | Prevent the selection of one's own organization when completing compliance and exclusion reports |1|](https://trello.com/c/sslrUM28/2093-reports-prevent-the-selection-of-ones-own-organization-when-completing-compliance-and-exclusion-reports-1)

PR #1619, opened 2019-11-15
Organization Autocomplete no longer includes own's organization

amichard:
#1618 

Changelog:
- Added reading of "request" in the autocomplete
- Exclude the user's own organization for the autocomplete

### DIFF
--- a/backend/api/services/Autocomplete.py
+++ b/backend/api/services/Autocomplete.py
@@ -152,7 +152,7 @@ class Autocomplete:
     }
 
     @staticmethod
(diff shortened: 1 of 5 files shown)


amichard:
[Reports | Prevent the selection of one's own organization when completing compliance and exclusion reports |1|](https://trello.com/c/sslrUM28/2093-reports-prevent-the-selection-of-ones-own-organization-when-completing-compliance-and-exclusion-reports-1)

amichard:
Changed so that it's using a parameter instead of just assuming we want to exclude the user's organization

Issue #1620, opened 2019-11-15
Notifications | Mark all as read |2|


amichard:
**Task:**

Implement mark all as read functionality for notifications (on the main notifications page)

https://trello.com/c/IuygNJrF/1579-notifications-mark-all-as-read-2

amichard:
![](https://github.trello.services/images/mini-trello-icon.png) [Notifications | Mark all as read |2|](https://trello.com/c/IuygNJrF/1579-notifications-mark-all-as-read-2)

PR #1621, opened 2019-11-15
Added Mark all as Read action

amichard:
#1620 

Changelog:
- Mark all as read button should show up in the "dropdown"
- Added condition in the viewset to select all unread for the user
- Added condition to check and make sure its for the right user


amichard:
[Notifications | Mark all as read |2|](https://trello.com/c/IuygNJrF/1579-notifications-mark-all-as-read-2)

Issue #1623, opened 2019-11-19
Compliance Report | User Activity history for compliance and exclusion reporting |3|


amichard:
**Task:**
This card is to either create a second User Activity table for history tracking for Compliance/Exclusion Reporting or to amend the existing User Activity table to include history tracking for compliance/exclusion reporting. 

**Location in TFRS:**

- BCeID: Organization Details > User > User Activity
- IDIR: 
 - (1) Administration > User Activity
 - (2) Administration > User > User Activity

**Information to include:**

- User
- Action: Submitted, Recommended Acceptance, Recommended Rejection, Accepted, Rejected, Requested Supplemental
- Report Type: Compliance or Exclusion Report
- Organization: the organization in which the report pertains to
- Timestamp

**Adjustment to Existing User Activity Table:**
We also want to change one of the labels on the existing User Activity tables. The label "fuel supplier" should be changed to "organization"

- see attached screenshot

https://trello.com/c/w59xWI9F/1706-compliance-report-user-activity-history-for-compliance-and-exclusion-reporting-3

PR #1624, opened 2019-11-20
Compliance Report History

amichard:
#1623 

The code won't win any beauty pageant, but it is working.

Changelog:
- Added compliance reporting as part of the user activity
- Added get_compliance_report_history to User model

### DIFF
--- a/backend/api/models/User.py
+++ b/backend/api/models/User.py
@@ -22,13 +22,14 @@
 """
 
 from django.db import models
-from django.db.models import Q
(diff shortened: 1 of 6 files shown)


amichard:
[Compliance Report | User Activity history for compliance and exclusion reporting |3|](https://trello.com/c/w59xWI9F/1706-compliance-report-user-activity-history-for-compliance-and-exclusion-reporting-3)

PR #1626, opened 2019-11-21
Fuel code version autocomplete now ignores cancelled fuel codes

amichard:
Changelog:
- Should now ignore cancelled fuel codes when automatically generating the next fuel code version
- New migration so unique constraint in the database takes into account the status
- Updated the validator in the serializer so when it does the check it checks for Cancelled. Otherwise, you will be able to add one draft and one submitted using the same fuel code version


amichard:
Closing for now. Needs a migration change.

PR #1631, opened 2019-11-25
Supplemental Report History Updates

plasticviking:
# Changelog
- Update status history component to better show position of changelogs
- Fix a defect in supplemental report acceptance reduction calculations
- Update compliance assessment tab to show all transactions in a report's history
- Remove report type, add Pacific timestamp to compliance report list
- Adjust language for supplemental note
- Fix tooltip z-index
- Remove some vestiges of old snapshot-related code


frontend/src/compliance_reporting/components/ComplianceReportingTable.js
```diff
@@ -123,18 +78,16 @@ class ComplianceReportingTable extends Component {
       id: 'supplemental-status',
       minWidth: 75
     }, {
-      accessor: item => (item.updateTimestamp ? moment(item.updateTimestamp).format('YYYY-MM-DD') : '-'),
+      accessor: item => (item.sortDate ? item.sortDate : null),
```
amichard: This will break the filtering. Ideally, the code for the new format is here in the accessor instead of Cell

plasticviking: Resolved in a way that filter and sort both work

PR #1680, opened 2020-09-24
Added new config for backend url

amichard:
This should allow us to specify a backend url instead of just expecting it to use the frontend url with /api at the end


kuanfandevops (approved):
looks good

PR #1684, opened 2020-10-01
Added a fallback filename for downloading

amichard:
Changelog:
- Added a logic to generate a filename if the content-disposition can't be retrieved


kuanfandevops (approved):
looks good

PR #1737, opened 2021-03-31
Max Credit Offset Calculation Fix + Credit Offset Validation

amichard:
Changelog:
- Credit balance was off as penalty was being deducted before it checks for available balance, so it's fixed to go after
- Added validation so people can't paste in a higher value than their available credits


emi-hi (approved):

kuanfandevops (approved):
looks good

PR #1738, opened 2021-03-31
Fixed validation

amichard:
Changelog:
- Moving up the validation so it doesn't count the current report twice


emi-hi (approved):

kuanfandevops (approved):
looks good

PR #1740, opened 2021-04-14
Fixed internal comments showing up for supplier spreadsheet

amichard:
Changelog:
- Added a user check to the spreadsheet builder so we can check for permission on whether they can view internal comments


kuanfandevops (approved):
looks good

PR #1742, opened 2021-04-15
Applying the comment privacy fix to 1.6.24

amichard:
Changelog:
- Added a user check to the spreadsheet builder so we can check for permission on whether they can view internal comments


kuanfandevops (approved):
Looks good

PR #1796, opened 2022-01-05
Fixed all the issues caused by upgrades

amichard:
I didn't realize that I had to use python3 to see if the new packages work
Changelog:
- Fixed issues with the updated packages


kuanfandevops (approved):
looks good

PR #1797, opened 2022-01-06
File submission role should now show up even for inactive organizations

amichard:
Changelog:
- Updated permission_required function to method_decorator
- Removed condition for file submission role for inactive organizations


kuanfandevops (approved):
looks good

PR #1802, opened 2022-01-20
Fixed the address not loading properly with the new packages

amichard:
Changelog:
- The new restframework version changed how it interprets underscores. We were expecting address_line_1 to become AddressLine_1, but it became AddressLine1


kuanfandevops (approved):
Looks good

PR #1804, opened 2022-01-20
Fixed an issue during creation

amichard:
Changelog:
- Previous changes broke the creation process. this should fix it


kuanfandevops (approved):
approved

PR #1938, opened 2022-12-19
TFRS-517 Fuel code dropdown issue fixed

vibhiquartech:


AlexZorkin (changes requested):
Removing this id would return fuel_codes not associated with the appropriate approved_fuel_type. Please re-review task 517 and come up with a different approach.

PR #2059, opened 2023-01-30
fix: tfrs-2050 - autocompletion

tim738745:
@AlexZorkin I noticed that the problem seemed to be that a new component from react-search-autocomplete was used to replace the original component from react-autocomplete, but it looks like the necessary changes to complete this switch were not made; I think if we do want to switch to using react-search-autocomplete, that should be another card? For now I just reverted back to using react-autocomplete, so that the feature works.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -49,7 +49,6 @@
     "react-redux-toastr": "7.6.10",
     "react-router": "^6.4.2",
     "react-router-dom": "^6.4.2",
-    "react-search-autocomplete": "^8.0.0",
     "react-autocomplete": "^1.8.1",
     "react-table": "6.10.3",
     "react-text-mask": "^5.5.0",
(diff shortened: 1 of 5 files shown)


AlexZorkin:
Gotcha, thanks Tim. Yes this was a missed fix during the code refactoring. If react-autocomplete works with the latest dependancies then we can just use that. Can you remove reference to react-search-autocomplete please.

tim738745:
@AlexZorkin Found that react-search-autocomplete also replaced "react-bootstrap-autosuggest", but we can't go back to using that since react 15 is one of its peer dependencies; I'm going to try to use react-autocomplete to replace it instead.

tim738745:
@AlexZorkin okay I removed react-search-autocomplete and replaced usage of react-bootstrap-autosuggest with our custom AutocompletedInput component (which uses react-autocomplete); should be okay now.

AlexZorkin (approved):

PR #2078, opened 2023-02-08
TFRS - Task - Dashboard compliance reports load time#2037 

jig-patel:
dashboard slow temp FE fix.


AlexZorkin (changes requested):
director_component_fix
Please update the styling to show the loading indicator inline with the text, rather than above.

jig-patel:
@AlexZorkin Do you have example on how it should look ?

AlexZorkin (approved):

PR #2118, opened 2023-02-22
TFRS-2012- Removed react/prop-types

vibhiquartech:


AlexZorkin (changes requested):
This is a good start but there are way more components that need propType updates still. just run npm run lint to see all the components with issues.

PR #2148, opened 2023-03-02
TFRS-2103- Added new penalty for carbon

vibhiquartech:


AlexZorkin (changes requested):
Looking good, we need to include a few unit tests to cover this new logic. Please add a backend test the covers the year and checks for the correct line 28 value. Please add a frontend test that tests the ScheduleSummaryPart3 component with props.period being different years.

frontend/src/schedule_summary/__tests__/ScheduleSummaryPart3.test.js
```diff
@@ -0,0 +1,475 @@
+import React from 'react'
+import ScheduleSummaryPart3 from '../ScheduleSummaryPart3';
+import FontAwesomeIcon from '@fortawesome/react-fontawesome'
+import { numericColumn, numericInput, totalViewer } from '../../compliance_reporting/components/Columns'
+import Tooltip from '../../app/components/Tooltip'
+describe('ScheduleSummaryContainer',() => {
+    it("When period < 2023", () => {
+        const scheduleSummaryPart3 = new ScheduleSummaryPart3(2022);
+        const expectedValue = [
```
AlexZorkin, line 9: This expected value doesn't need to include the whole part3 object. We're looking to test just that the title value has changed. Please update this to something like the following:
`    expect(result[SCHEDULE_SUMMARY.LINE_28][1].value.props.children).toEqual([
      'Line 28 ',
      <Tooltip // eslint-disable-line react/jsx-key
        className="info"
        show
        title="This line displays the penalty payable based on the information provided and is calculated using the $200 per outstanding debit non-compliance penalty."
      >
        <FontAwesomeIcon icon="info-circle" />
      </Tooltip>
    ])`

frontend/src/schedule_summary/__tests__/ScheduleSummaryPart3.test.js
```diff
@@ -0,0 +1,475 @@
...
+                className: 'total',
+                readOnly: true,
+                value: '$CAD'
+              }]
+          ];
+          expect(scheduleSummaryPart3).toEqual(expectedValue);
+    });
+    it("When period > 2022", () => {
+        const scheduleSummaryPart3 = new ScheduleSummaryPart3(2023);
+        const expectedValue = [
```
AlexZorkin, line 243: same as above, but with different target text.

AlexZorkin (changes requested):

PR #2155, opened 2023-03-08
TFRS-2116 - Added filtering & sorting to the Compliance and Exclusion Reports page for columns -Original & Current Status, Last Updated On

vibhiquartech:


backend/api/viewsets/ComplianceReport.py
```diff
@@ -241,7 +300,7 @@ def get_latest_supplemental_reports(self):
         latest_supplementals = ComplianceReport.objects.raw("""
             select distinct on (p.id) c.*
             from compliance_report p 
-              left join compliance_report c on p.id = c.supplements_id
+            left join compliance_report c on p.id = c.id
```
AlexZorkin, line 303: This seems odd. It's joining on itself and would only ever return one record. What was the intended purpose of this change?

AlexZorkin, line 303: This seems odd. It's joining on itself and would only ever return one record. What was the intended purpose of this change?

AlexZorkin (changes requested):
just one question, otherwise looks good.

backend/api/viewsets/ComplianceReport.py
vibhiquartech: Alex, I forgot to revert the changes while testing supplemental status column. I have reverted back the changes. Thanks for bringing that up

AlexZorkin (approved):

PR #2184, opened 2023-03-18
BC attorney address add and unit testing

jig-patel:
BC attorney address add and unit testing


frontend/src/organizations/OrganizationEditContainer.js
```diff
@@ -3,229 +3,272 @@
...
+    this._handleInputChange = this._handleInputChange.bind(this);
+    this._handleCreate = this._handleCreate.bind(this);
+    this._handleUpdate = this._handleUpdate.bind(this);
   }
 
-  componentDidMount () {
-    if (this.props.mode === 'add') {
-      return
+  componentDidMount() {
+    if (this.props.mode === "add") {
```
AlexZorkin, line 59: We are using JS Standard linting which doesn't use semi-colons. Try running npm run lint to see any required changes.

backend/api/migrations/0204_auto_20230316_2115.py
```diff
@@ -0,0 +1,21 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('api', '0203_auto_20230312_0823'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
AlexZorkin, line 13: are we sure there isn't any information we're losing with this removal?

AlexZorkin (changes requested):

backend/api/migrations/0204_auto_20230316_2115.py
jig-patel: We would need to manually update DB to append Address_line_3 data with Address_line_2 and county content with Province unless there is better way to do it. In a local i did update DB manually with query to make sure we don't lose data.

frontend/src/organizations/OrganizationEditContainer.js
jig-patel: i did ran "npm run lint" and it didn't complain about semi-colon.

backend/api/migrations/0204_auto_20230316_2115.py
kuanfandevops: @jig-patel please update the based branch to release-2.5.0 as 2.4.0 has been released to prod

jig-patel: @kuanfandevops thanks. will do

frontend/src/organizations/OrganizationEditContainer.js
jig-patel: i will take that previous statement back. npm run lint fixed semicolon and i have pushed that changes.

PR #2198, opened 2023-03-24
TFRS-2182- Removed All References to Keycloak in Project TFRS

vibhiquartech:
@kuanfandevops :this PR removes references to the keycloak service account from the deployment configs. Can you please have a look to check if everything looks okay? Thanks.


kuanfandevops:
@vibhiquartech looks like backend doesn't need any keycloak references?

PR #2210, opened 2023-03-30
TFRS-2130- Prevent Supplier from updating address their address info

vibhiquartech:


AlexZorkin (changes requested):
We don't want to remove edit functionality for everyone, just for suppliers. You will have to check whether a supplier is logged in or a government user.

PR #2226, opened 2023-04-05
Tfrs 2183

jig-patel:
Modal.js refactor to use react based component


AlexZorkin (changes requested):
Looking good, let's also remove any references to Jquery, in the package.json and in the .html. Clear out any other references as well.

jig-patel:
There are multiple other places where jquery is being used which is not related to Modal.js so removing reference of jquery from .html file would cause problem. for example SessionTimer.js is using jquery with different modal. bootstrap.js has jquery  references. to clean up all reference of jquery would required ticket on its own.

PR #2282, opened 2023-05-02
fix: tfrs-2127 - use external username

tim738745:


backend/api/services/SpreadSheetBuilder.py
```diff
@@ -238,7 +239,17 @@ def add_users(self, fuel_supplier_users):
             worksheet.write(row_index, 0, user.last_name)
             worksheet.write(row_index, 1, user.first_name)
             worksheet.write(row_index, 2, user.email)
-            worksheet.write(row_index, 3, user.display_name)
+
+            keycloak_user_id = user.keycloak_user_id
```
AlexZorkin, line 243: Using UserLoginHistory does work but it may be more reliable to use the UserCreationRequest model to get the external_username. Did you make a judgement call between these two models?

tim738745: @AlexZorkin Yeah I'm working with data from the test environment, and I saw 91 records in user_creation_request and 115 records in users, so my reasoning was that not every user is going to have a user_creation_request, which is the same case for user_login_history, so they seemed equivalent to me. But I'm actually not 100% clear on how user_creation_requests work, so if you think that's more reliable, I can change it to that.

tim738745: Is it that every user that has a keycloak_user_id is guaranteed to be associated with a user_creation_request?

tim738745: I just tried it with User_creation_request and I got way more external usernames, so I'll use that. Thanks!

PR #2336, opened 2023-05-26
Compliance Report slow loading time fix

prv-proton:
Compliance Report Model changes to include latest_report and root_report information To migrate the data to these new columns
Using select_related and prefetch_related wherever necessary Making use of the new columns in data retrieval


backend/api/serializers/ComplianceReport.py
```diff
@@ -209,8 +210,14 @@ class ComplianceReportWorkflowStateSerializer(serializers.ModelSerializer):
     )
 
     def should_show(self, field_name, value):
-        user = self.context['request'].user \
-            if 'request' in self.context else None
+        user = cache.get("user")
```
AlexZorkin, line 213: This use of caching a user is a security concern because the user would be accessible globally and could be picked up by a different user using the system. Please revert these bits of caching code and get the user from the request context.

backend/api/serializers/ComplianceReport.py
```diff
@@ -262,12 +266,29 @@ def get_display_name(self, obj):
         return obj.generated_nickname
 
     def get_group_id(self, obj):
-        user = self.context['request'].user
+        user = cache.get("user")
```
AlexZorkin, line 269: please revert this caching of user as related to above

backend/api/serializers/ComplianceReport.py
```diff
@@ -262,12 +266,29 @@ def get_display_name(self, obj):
...
+        ).values('supplements_id').distinct()
+
+        qs = ComplianceReport.objects.filter(
+            Q(id__in=Subquery(distinct_supplements_ids)) | Q(id=obj.latest_report.id)
+        ).order_by('create_timestamp').select_related('type').select_related('status')
+
+        if obj.latest_report.id == obj.id:
+            qs = qs.filter(~Q(id=obj.latest_report.id))
+            
+        gov_org = cache.get("organization-type-1")
```
AlexZorkin, line 288: This is a good use of caching because it can be globally available and reused, avoiding the db call each time. We can use this principle elsewhere in the app for more caching situations.

backend/api/models/User.py
```diff
@@ -96,7 +96,11 @@ def roles(self):
         """
         Roles applied to the User
         """
-        return Role.objects.filter(user_roles__user_id=self.id)
+        role = cache.get(f'role-user-{self.id}')
```
AlexZorkin, line 99: I don't think we should cache user roles because if an admin changes the assigned roles, they won't be updated immediately in the cache and could cause confusion.

backend/api/migrations/0004_migrate_compliance_report.py
```diff
@@ -0,0 +1,58 @@
...
+                        not current.status.fuel_supplier_status_id == "Deleted":
+                    latest = current
+                    i += 1
+                id_traversal[current_id] = i
+                for descendant in current.supplemental_reports.order_by('create_timestamp').all():
+                    to_visit.append(descendant.id)
+
+            for compliance_id, traversal in id_traversal.items():
+                ComplianceReport.objects.filter(id=int(compliance_id)) \
+                    .update(latest_report=latest, root_report=root, traversal=traversal)
```
AlexZorkin, line 45: Is this the only place the the traversal value is being updated? Should it be set in the super save callback?

backend/api/models/ComplianceReport.py
```diff
@@ -226,49 +247,25 @@ class ComplianceReport(Auditable):
         db_comment='An explanatory note required when submitting a supplemental report'
     )
 
+    def save(self, *args, **kwargs):
+        if self.supplements:
+            root = self.supplements.root_report or self.supplements
+            self.root_report = root
+            self.latest_report = self
```
AlexZorkin, line 254: Should we be setting the traversal value here as well?

backend/api/models/User.py
```diff
@@ -162,7 +166,11 @@ def is_government_user(self):
         """
         Does this user have a government role?
         """
-        if self.roles.filter(Q(is_government_role=True)):
+        is_gov_user = cache.get(f'role-user-{self.id}-gov-role-true')
```
AlexZorkin, line 169: again i don't think we want to cache this value in case roles change. Even though a gov user should only ever be the user with gov roles, i still think we should avoid caching this value.

backend/api/viewsets/ComplianceReport.py
```diff
@@ -406,12 +409,7 @@ def filter_draft(self, latest_supplemental):
 
     def get_latest_supplemental_reports(self):
         latest_supplementals = ComplianceReport.objects.filter(id__in=RawSQL("""
```
AlexZorkin, line 411: Do we still need to use a raw query here or could we use just a Django query? is the raw query faster in this case?

AlexZorkin (changes requested):

backend/api/migrations/0004_migrate_compliance_report.py
prv-proton: Yes, will include in the save method which was missed out.

AlexZorkin (approved):

PR #2509, opened 2023-08-10
Fix: updated category logic on backend for credit trades, approved category text - 2472 2503 2470

AlexZorkin:


jig-patel (changes requested):
Category check box is still being displayed in Rescinded state.

jig-patel:
xls download sheet, category column is empty. Should display proper category for approved transactions.

PR #2524, opened 2023-08-16
Feat: updated organization label column for status to registered - 2488 2490 2483

AlexZorkin:


frontend/src/organizations/components/OrganizationsTable.js
```diff
@@ -40,10 +40,18 @@ const OrganizationsTable = (props) => {
...
+    accessor: item => {
+      const orgStatus = Object.values(ORGANIZATION_STATUSES)
+        .find(element => element.id === item.status)
+      if (orgStatus.description === 'Active') {
+        return 'Yes'
+      } else if (orgStatus.description === 'Inactive') {
+        return 'No'
+      }
+      return orgStatus.description // default to the actual description if not Active/Inactive
+    },
```
prv-proton, line 52: In the filter it is expecting to enter entire word "Yes/No", instead of filtering based on each character being entered.

prv-proton (changes requested):

PR #2529, opened 2023-08-16
System converting credits to compliance units for 2023 prior reports

prv-proton:
System converting credits to compliance units for 2023 prior compliance reports


frontend/src/schedule_summary/ScheduleSummaryPart3.js
jig-patel, line 1: there is extra curly braces and return statement for part 3 need to move one curly brace up.

AlexZorkin (approved):

PR #2530, opened 2023-08-16
removing buy option from credit transfer

jig-patel:


frontend/src/credit_transfers/components/CreditTransferFormDetails.js
```diff
@@ -27,21 +27,7 @@ class CreditTransferFormDetails extends Component {
...
-              >
-                <option value="" />
-                <option value="1">Sell</option>
-                <option value="2">Buy</option>
-              </select>
-            </div>
-          }
+          <span> sell </span>
           {this.props.loggedInUser.organization.actionsTypeDisplay !== 'Buy And Sell' &&
             <span> sell </span>
```
AlexZorkin, line 32: Is this <span> sell </span> now duplicated or redundant?

jig-patel: its redundant. by default it does sell transaction.

AlexZorkin: I mean this line here:  {this.props.loggedInUser.organization.actionsTypeDisplay !== 'Buy And Sell' &&
will that ever happen? Then we would see "sell sell"?

jig-patel: oh. yes. if organization is Inactive then it will happen and we could see "sell sell". i will fix it since in main release we still want to have supplier to allow sell credit if they are inactive.

AlexZorkin (approved):

PR #2568, opened 2023-08-29
Fix: Credit Transaction Excel Effective Date - 2536

AlexZorkin:
Made all credit transfer types show effective_date as either trade_effective_date or last_updated_at.

### DIFF
--- a/backend/api/services/SpreadSheetBuilder.py
+++ b/backend/api/services/SpreadSheetBuilder.py
@@ -150,11 +150,8 @@ class SpreadSheetBuilder(object):
             else:
                 worksheet.write(row_index, 8, credit_trade.status.friendly_name)
 
-            # If the trade has an effective date, write it to the worksheet.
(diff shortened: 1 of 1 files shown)


jig-patel:
@AlexZorkin is Effective date should be seen for approved transaction only or it should be visible  for other state as well (e.g signed, declined, rescinded ?) Currently its visible for all except Declined.

AlexZorkin:
> @AlexZorkin is Effective date should be seen for approved transaction only or it should be visible for other state as well (e.g signed, declined, rescinded ?) Currently its visible for all except Declined.

If a credit transfer has those statuses other than 'Approved' then they should not show an effective date. With this pr it fixes that issue by only allowing Approved credit transfers with an effective date, as well as any transfers of type "Credit Reduction", "Credit Validation"

jig-patel (approved):

PR #2580, opened 2023-09-01
Remove Actions and Last Transaction column from Organizations table

jig-patel:


AlexZorkin:
This ticket also requires the removal of the actions column from the organizations excel download. In the SpreadsheetBuilder.py you'll have to update the add_fuel_suppliers method to remove this column when called by the organization viewset. The CreditTrade download also uses this method and currently we still want to include the actions column in that download.

AlexZorkin (changes requested):
This ticket also requires the removal of the actions column from the organizations excel download. In the SpreadsheetBuilder.py you'll have to update the add_fuel_suppliers method to remove this column when called by the organization viewset. The CreditTrade download also uses this method and currently we still want to include the actions column in that download.

jig-patel:
Thansk @AlexZorkin .made changes to exclude actions from organization but keep it when downloading from creditTrade

AlexZorkin (approved):

PR #2597, opened 2023-09-11
TFRS - Credit Transfer Rescinded and Declined Format Issues#2523

jig-patel:


backend/api/viewsets/CreditTrade.py
```diff
@@ -146,7 +146,7 @@ def perform_update(self, serializer):
         credit_trade = serializer.save()
         # we only want to create history and send notifications
         # when a status change occurs
-        if previous_state.status != credit_trade.status:
+        if previous_state.status != credit_trade.status or credit_trade.is_rescinded:
```
AlexZorkin, line 149: This is a bit concerning because if an update is performed for a different reason on a rescinded credit transfer, then a new duplicated history item will be created. When a credit transfer is rescinded, is the perform_update not used?

AlexZorkin (changes requested):

backend/api/viewsets/CreditTrade.py
jig-patel: i thought once report is rescinded, there is no option to update it. since when we click on "rescind" button on UI we are passing {is_rescinded: 'True'} instead of {status:11}, perform_update method "if previous_state.status != credit_trade.status" becomes false and it doesn't create history.
i just logged in as gov user and supplier and see after report is rescinded, there is no edit button to update it.

AlexZorkin: Ideally just as a safeguard for the future we add a check for any existing rescinded history records. There should only ever be one rescinded history record.

Issue #2469, opened 2023-07-26
TFRS - Adjust credit transfer agreement date error message


justin-lepitzki:
**Describe the task**
Adjust the error message that appears in the UI when the agreement date field within a draft credit transfer is left blank. This message appears when the user clicks on the Propose button. 

- Current message: "Error! Date has wrong format. Use one of these formats instead: YYYY-MM-DD."

- Revised message: "Error! Please enter a valid date in the Agreement Date field: YYYY-MM-DD."

**Purpose**
More descriptive of the issue to help the user understand what went wrong. 

**Acceptance Criteria**
- [ ] When a user proposes a credit transfer with an empty agreement date, they should see the revised error message.

**Screenshot:**
![Screenshot 2023-07-25 CTP agreement date error.png](https://images.zenhubusercontent.com/644963ac99f1d06cf29952ef/5ded287c-7580-45c6-92f8-2f6f6db29401)

kcabhar:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Team-Zelda-637d5bc12e6201001055b745/issues/bcgov/tfrs/2469?planning-poker) with Zenhub @AlexZorkin @jig-patel @prv-proton

PR #2706, opened 2023-10-31
fix: adjust credit transfer agreement date error message (#2469)

hvalioll:
Improve the error message shown when a user tries to propose a credit transfer with an empty Agreement Date field.

Previous message: 'Error! Date has wrong format. Use one of these formats instead: YYYY-MM-DD.'
New message: 'Error! Please enter a valid date in the Agreement Date field: YYYY-MM-DD.'

### DIFF
--- a/backend/api/serializers/CreditTrade.py
+++ b/backend/api/serializers/CreditTrade.py
@@ -281,6 +281,11 @@ class CreditTradeCreateSerializer(serializers.ModelSerializer):
                                       "the transaction."
                 }
             },
+            'date_of_written_agreement': {
+                'error_messages': {
+                    'invalid': "Please enter a valid date in the Agreement Date field: YYYY-MM-DD."
+                }
+            },
             'trade_effective_date': {
                 'required': False
             }


backend/api/serializers/CreditTrade.py
```diff
@@ -281,6 +281,11 @@ class Meta:
                                       "the transaction."
                 }
             },
+            'date_of_written_agreement': {
+                'error_messages': {
+                    'invalid': "Please enter a valid date in the Agreement Date field: YYYY-MM-DD."
```
prv-proton, line 286: Since in the UI we're inputting the data in the format of mm/dd/yyyy, does YYYY-MM-DD format makes sense?

hvalioll: Thank you, Prashanth, for reviewing my PR. From what I can see, the user inputs the agreement date in the YYYY-MM-DD format. Could you please take a look at the attached screenshots to confirm whether we need to make any changes? I have attached two screenshots: one before entering the date and one after.

Before entering the date:
Screen Shot 2023-10-31 at 10 31 43 AM

After entering the date:
Screen Shot 2023-10-31 at 10 31 50 AM

prv-proton: Yes, you're right. Since in my system I had chosen the date format to be in US format. it was showing as below
image

Thanks,

prv-proton (approved):

backend/api/serializers/CreditTrade.py
hvalioll: Thanks for the info! We can discuss this in the sprint meeting tomorrow and create a new ticket if needed. I think if the user has the option to change the date format, the error message should adjust accordingly.

AlexZorkin (approved):

PR #2709, opened 2023-11-01
fix: removed 2023 option from compliance report dropdown

### LINKED ISSUE #2553: TFRS - Prevent fuel suppliers from creating draft compliance reports for 2023
**Describe the task**
Remove the ability for BCeID users to create draft compliance reports for 2023. We have already prevented the creation of draft compliance & exclusion reports for 2024 and beyond. 

**Purpose**
The new Act applies compliance units to the 2023 compliance period which we are currently developing in TFRS. A fuel supplier can create a 2023 report right now, but it will be incorrect as it will use the old credits/debits system. Therefore, we want to prevent users from creating new 2023 reports until we have released the new compliance units feature for 2023 reporting. 

**Acceptance Criteria**
- [ ] Remove 2023 option from dropdown  

**Screenshot:**

kevin-hashimoto:
removes the option to select 2023 as a compliance reporting period

closes #2553

### DIFF
--- a/frontend/src/compliance_reporting/components/ComplianceReportingPage.js
+++ b/frontend/src/compliance_reporting/components/ComplianceReportingPage.js
@@ -334,7 +334,7 @@ const ComplianceReportingPage = (props) => {
             {
               Array.from({ length: new Date().getFullYear() - 2017 }, (_, index) => {
                 const year = new Date().getFullYear() - index + 1
-                if (year === 2023) return null
+                if (year >= 2023) return null
                 return (
                 <option key={year} value={year}>
                   {year}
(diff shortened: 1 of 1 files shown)


frontend/src/compliance_reporting/components/ComplianceReportingPage.js
```diff
@@ -334,6 +334,7 @@ const ComplianceReportingPage = (props) => {
             {
               Array.from({ length: new Date().getFullYear() - 2017 }, (_, index) => {
                 const year = new Date().getFullYear() - index + 1
+                if (year === 2023) return null
```
AlexZorkin, line 337: Let's make this null out any years greater than or equal to 2023. I know the ticket says we've already handled 2024 and above but it makes more sense just to limit the year values here.

AlexZorkin (changes requested):

frontend/src/compliance_reporting/components/ComplianceReportingPage.js
kevin-hashimoto: @AlexZorkin resolved

AlexZorkin (approved):

Issue #2664, opened 2023-10-11
TFRS - File submission filtering from dashboard link not working - IDIR only 


justin-lepitzki:
**Describe the Bug**
The dashboard links (x2) to in progress file submissions are not working. 

**Expected Behaviour**
When an IDIR user clicks on one of the two links (see attached screenshot), they should be taken to the File Submissions page with status filters applied to the main file submissions react table.

**Actual Behaviour**
When an IDIR user clicks on one of the two links, they are taken to the File Submissions page with no filters applied (i.e., full display of file submissions regardless of status).

**Steps To Reproduce**
Steps to reproduce the behaviour:
User/Role: IDIR; non-director role
1. Go to the Dashboard 
2. Click on one of the two links, see screenshot here: [Private Zenhub Image](https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBN3dvQVE9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--09cd671fdd88626fd6a9b7d73c50ba3cc0620e7a/Screenshot%202023-10-11%20FS.png)
3. See error (unfiltered table)

AlexZorkin:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Low-Carbon-Fuel-Standard-644c2db3be098a001977ad0b/issues/bcgov/tfrs/2664?planning-poker) with Zenhub @jig-patel @prv-proton

PR #2710, opened 2023-11-01
TFRS - File submission filtering from dashboard link not working - IDIR only #2664

jig-patel:
TFRS - File submission filtering from dashboard link not working - IDIR only #2664


AlexZorkin (approved):

AlexZorkin:
This needs to be merged into the january release branch instead of main branch

AlexZorkin (changes requested):
need to point this to the January release branch

PR #2722, opened 2023-11-03
fix: updated transfer labels for BCeID users according to new Act

### LINKED ISSUE #2690: TFRS - a few more new Act label changes
**Describe the task**
A couple more new Act label changes for the Transfer feature.

**Change 1:**
- New transfer page (BCeID only): change "Select a Fuel Supplier" to "Select an Organization"
- 

**Change 2:**
- 

**Purpose**
To align with the framing and descriptions in the new Act 

**Acceptance Criteria**
- [ ] first
- [ ] second
- [ ] third

**Additional context**
- Add any other context about the task here.
- Or here

hamed-valiollahi:
This commit makes the following updates to the New Transfer page for BCeID users, ensuring alignment with the new Act:

- Change the dropdown label from 'Select a Fuel Supplier' to 'Select an Organization'.

- The signing authority declaration statement is now updated to: 'I confirm that records evidencing each matter reported under section 17 of the Low Carbon Fuel (General) Regulation are available on request.'

Additionally, a database migration is added to apply the signing authority declaration statement label change within the existing record.

Closes #2690

### DIFF
--- a/backend/api/migrations/0012_update_signing_authority_declaration_statement.py
+++ b/backend/api/migrations/0012_update_signing_authority_declaration_statement.py
@@ -22,7 +22,8 @@ def update_sign_auth_assertion(apps, schema_editor):
         )
         assertion.save()
     except signing_authority_assertion.DoesNotExist:
-        logging.error('Failed to update SigningAuthorityAssertion: No entry found with id "1".')
+        logging.warning('Failed to update SigningAuthorityAssertion: No entry found with id "1".')
+        raise
 
 class Migration(migrations.Migration):
     """
--- a/backend/api/migrations/0012_update_signing_authority_declaration_statement.py
+++ b/backend/api/migrations/0012_update_signing_authority_declaration_statement.py
@@ -1,0 +1,37 @@
+import logging
+from django.db import migrations
+
+def update_sign_auth_assertion(apps, schema_editor):
+    """
+    Updates the signing authority declaration statement
+
+    Previous label:
+    "I confirm that records evidencing each matter reported under section 11.11 (2) of the 
+    Regulation are available on request."
+
+    New label:
+    "I confirm that records evidencing each matter reported under section 17 of the Low Carbon
+    Fuel (General) Regulation are available on request."
+    """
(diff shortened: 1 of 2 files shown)


backend/api/migrations/0012_update_signing_authority_declaration_statement.py
```diff
@@ -0,0 +1,38 @@
...
+    "I confirm that records evidencing each matter reported under section 11.11 (2) of the 
+    Regulation are available on request."
+
+    New label:
+    "I confirm that records evidencing each matter reported under section 17 of the Low Carbon
+    Fuel (General) Regulation are available on request."
+    """
+    signing_authority_assertion = apps.get_model('api', 'SigningAuthorityAssertion')
+    try:
+        assertion = signing_authority_assertion.objects.get(id=1)
```
AlexZorkin, line 18: Are we certain that id=1 is the correct assertion here? Is there a more consistent way to get the correct assertion across environments?

AlexZorkin (changes requested):

backend/api/migrations/0012_update_signing_authority_declaration_statement.py
hamed-valiollahi: You are right, Alex! Using record IDs isn't a good practice because they can change in different environments. I was unable to find any unique names for the `SigningAuthorityAssertion` records in the code. Would you like me to add a unique name to each record? Then I could update this migration file to use these unique names instead of the automatically created IDs.

Another solution would be to find the records by their text (description). This isn't ideal, but it is still better than updating the records by their IDs.

AlexZorkin: hmm yeah a bit tricky if there isn't already a different unique field, and selecting by long form text isn't idea. Let's just continue with your id solution but please confirm that each environment uses those ids for our target records.

hamed-valiollahi: I can confirm that all the environments (test, development, and production) use the same IDs for our target records.

PR #2754, opened 2023-12-11
fix: attorny address if null/empty

kevin-hashimoto:
- fixes an issue with attorney address appearing empty as the address checker checks for any fields that are not empty. the prefilled fields trigger the check and displays an empty attorney address in the ui
- created another similar check that omits the prefilled fields. Seems like the previous checker is still used somewhere so created a new one to not interfere

### DIFF
--- a/frontend/src/utils/functions.js
+++ b/frontend/src/utils/functions.js
@@ -256,15 +256,9 @@ const attornyAddressCheck = (address) => {
     if (address.attorneyCity) {
       return true
     }
-    // if (address.attorneyProvince) {
-    //   return true
-    // }
     if (address.attorneyPostalCode) {
       return true
     }
-    // if (address.attorneyCountry) {
-    //   return true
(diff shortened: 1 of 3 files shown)


frontend/src/utils/functions.js
```diff
@@ -242,7 +242,34 @@ const atLeastOneAttorneyAddressFieldExists = (address) => {
...
+    if (address.attorneyStreetAddress) {
+      return true
+    }
+    if (address.attorneyAddressOther) {
+      return true
+    }
+    if (address.attorneyCity) {
+      return true
+    }
+    // if (address.attorneyProvince) {
```
AlexZorkin, line 259: Are these commented out fields needed?

kevin-hashimoto: removed

AlexZorkin (approved):

Issue #2769, opened 2024-01-06
TFRS - Issues with the effective date for transactions entered using the Historical Data Entry feature


justin-lepitzki:
**Background**
Transactions entered using the historical data entry (HDE) feature have three dates associated with them: 
- Effective date: a specific input field in the HDE form that describes when the transaction is effective (when it occurred)
- Recorded date: the date the transaction was added to the queue in the HDE feature
- Approved date: the date the transaction was committed (approved) in the HDE feature

The download as xls feature allows the user to download data directly from TFRS. This feature is available on the Transactions page. Downloading the transactions data into a spreadsheet includes a column called 'Effective Date' that is intended to display the effective date for the different approved transactions. 

**Describe the Bugs**
Bug #1: The effective date is being changed to a future date (tomorrow's date) regardless of what the user inputs in the Effective Date field within the Historical Data Entry (HDE) feature. For example, today is January 5, 2024; the user inputs a transaction into the HDE feature and indicates that the effective date is December 15, 2023. Upon committing the transaction, the effective date is changed to January 6, 2024 (which is tomorrow's date, a future date). The effective date should be December 15, 2023. 
- Bug #1: is impacting TFRS main release dev

Bug #2:  The date being displayed in the downloaded transactions sheet for transactions entered using the historical data entry feature is the "Approved date". This is not the correct date to display. It should be showing the Effective date, which is the date entered into the specific input field in the HDE form. 
- Bug #2 is impacting TFRS main release Prod, but not TFRS main release Dev
- Bug #2 is impacting TFRS January release Dev

**Impacted transactions:**

- Both of these bugs are only impacting transactions entered using the HDE feature. They are not impacting transactions approved within the credit transfer feature and/or the compliance reporting feature. 

**Steps To Reproduce**
Bug #1: 
1) Login to TFRS dev, main release as an IDIR user: https://tfrs-dev.apps.silver.devops.gov.bc.ca/
2) Click on Historical Data Entry within the Administration section of the dashboard
3) Enter in a new transaction with an effective date in the past (e.g., some date in 2023).
4) Add the transaction to the queue, note that the correct effective date is displayed in the queue
5) Commit the transaction by clicking on the Commit button
6) Navigate back to the Credit Transactions page and click on the transactions you just committed
7) See error in effective date

Bug #2: 
See transaction ID #1217 in TFRS January release, dev environment: https://tfrs-dev.apps.silver.devops.gov.bc.ca/credit_transactions/view/1217
[...]

PR #2799, opened 2024-01-14
Fix: Effective Date fix for HDE entries-2769

prv-proton:
#2769

### DIFF
--- a/backend/api/services/CreditTradeService.py
+++ b/backend/api/services/CreditTradeService.py
@@ -178,14 +178,16 @@ class CreditTradeService(object):
         Sets the Credit Transfer to Approved
         """
         status_approved = CreditTradeStatus.objects.get(status="Approved")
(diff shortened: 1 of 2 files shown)


backend/api/services/CreditTradeService.py
```diff
@@ -178,14 +178,16 @@ def approve(credit_trade, update_user=None, batch_process=False):
         Sets the Credit Transfer to Approved
         """
         status_approved = CreditTradeStatus.objects.get(status="Approved")
+        effective_date = credit_trade.trade_effective_date
```
AlexZorkin, line 181: just want to confirm that on historical data entry, is the trade_effective_date always set? or can it be null?

prv-proton: I checked in the prod table and many records had null value.

AlexZorkin (approved):

PR #2807, opened 2024-01-18
Feat: Merge January Candidate - 240117

AlexZorkin:


kuanfandevops:
@AlexZorkin The frontend/package.json should keep version as 2.14.0.
Here are the up to date app and databases connection info:
	Dev:
		tfrs code -> spilo
		jan code -> crunchy
	Test:
		tfrs code -> spilo
		jan code -> crunchy
	Prod:
		tfrs code -> crunchy
the import-data.sh may be need to be revised after the merge

AlexZorkin:
> There is a 0014_update_trade_effective_dates.py on release-2.14.0 already. It may have conflicts when run migrations.

thanks, i fixed the ordering now.

Issue #2814, opened 2024-01-23
Date/time stamp records twice in report history box when hitting "recommend for acceptance"


Grulin:
**Describe the Bug**
IDIR user with government analyst role double clicks "Recommend for Acceptance" button on assessment page of report and the date/time stamp records twice in report history box.

**Expected Behaviour**
Should only record date/time stamp once when "Recommend for Acceptance" button is pressed. 

**Implications**
Unnecessary date/time stamp that could cause confusion for recommended for acceptance date. 

**Steps To Reproduce**
Steps to reproduce the behaviour:
User/Role: IDIR Government Analyst
1. Go to Assessment page of a submitted compliance report
2. Click on "Recommend for Acceptance" twice quickly or if page does not immediately load
3. Scroll up to grey text box and see if date/time stamp was recorded twice.
<img width="1566" src="https://api.zenhub.com/attachedFiles/eyJfcmFpbHMiOnsibWVzc2FnZSI6IkJBaHBBekh5QWc9PSIsImV4cCI6bnVsbCwicHVyIjoiYmxvYl9pZCJ9fQ==--82370eb115cacbd491ece4a73ea5a348021accc4/Double%20time%20stamp.png" alt="Double time stamp.png" />

![Image](https://github.com/bcgov/tfrs/assets/130073308/c3211b51-5908-4a29-bb54-10b465ea6140)

prv-proton:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/Low-Carbon-Fuel-Standard-644c2db3be098a001977ad0b/issues/bcgov/tfrs/2814?planning-poker) with Zenhub @AlexZorkin @hamed-valiollahi @kevin-hashimoto @prv-proton

prv-proton:
Wasn't able to re-create the bug as the button click will create a pop-up confirmation window. But have added extra check to prevent any duplicate history records from creating.

PR #2818, opened 2024-01-26
Fix: TFRS - Date/time stamp records twice in report history box when hitting "recommend for acceptance" #2814

prv-proton:
Fix: Date/time stamp records twice in report history box when hitting "recommend for acceptance"
#2814

### DIFF
--- a/backend/api/services/ComplianceReportService.py
+++ b/backend/api/services/ComplianceReportService.py
@@ -196,6 +196,16 @@ class ComplianceReportService(object):
             else:
                 role_id = user.roles.first().id
 
+        history = ComplianceReportHistory.objects.select_related('status').filter(
+            Q(compliance_report_id=compliance_report.id) &
+            Q(status__fuel_supplier_status__status=compliance_report.status.fuel_supplier_status_id) &
+            Q(status__analyst_status__status=compliance_report.status.analyst_status_id) &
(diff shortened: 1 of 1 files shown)


backend/api/services/ComplianceReportService.py
```diff
@@ -196,6 +196,16 @@ def create_history(compliance_report, is_new=False):
...
                 role_id = user.roles.first().id
 
+        history = ComplianceReportHistory.objects.filter(
+            Q(compliance_report_id=compliance_report.id) &
+            Q(status__fuel_supplier_status__status=compliance_report.status.fuel_supplier_status_id) &
+            Q(status__analyst_status__status=compliance_report.status.analyst_status_id) &
+            Q(status__manager_status__status=compliance_report.status.manager_status_id) &
+            Q(status__director_status__status=compliance_report.status.director_status_id))
+        # if a history record is already present with same status then don't create a new one.
+        if len(history) > 0:
```
AlexZorkin, line 206: This is good, but I think we should address the problem on the frontend as well. The button should be disabled after click, until a successful or failed response comes back. The view I think reloads anyway when the api call finishes so a disable should work.

AlexZorkin (changes requested):

AlexZorkin (approved):

PR #2919, opened 2024-11-06
Feat: TFRS: Add Redis Service for organization balance

prv-proton:
TFRS: Add Redis Service for Local Development
[#1078](https://github.com/bcgov/lcfs/issues/1078)

TFRS: Setup Redis Service and Call LCFS Cache for Available Balances
[#1074](https://github.com/bcgov/lcfs/issues/1074)


switch_application.sh
```diff
@@ -0,0 +1,96 @@
+#!/bin/bash
```
AlexZorkin, line 1: Can you add some more documentation here to explain the purpose and use case for this shell script please.

backend/api/app.py
```diff
@@ -51,11 +54,27 @@ def ready(self):
...
             except RuntimeError as error:
                 print('Startup checks failed. Not starting.')
                 print(error)
                 exit(-1)  # Django doesn't seem to do this automatically.
 
+    def seed_redis_cache(self):
+        """Seed initial data into Redis cache"""
+        try:
+            # Check if data has already been seeded to avoid duplicates
+            if not redis_cache.get('redis_seeded'):
```
AlexZorkin, line 68: We have to ensure this flag is set on the LCFS side as well when it populates its live cache. Please just confirm that is the case.

AlexZorkin (changes requested):
Looks good just a couple items to add. ty

switch_application.sh
prv-proton: updated documentation. 
I was hoping to not include the redis and rabbitmq into each other docker-compose files, instead we could just switch between tfrs and lcfs using this script while ensuring redis pod from LCFS container is active while tfrs application is running and rabbitmq from TFRS container is running while LCFS application is running.

backend/api/app.py
prv-proton: sure, will update LCFS repo as well.

switch_application.sh
AlexZorkin: That's great thank you.

AlexZorkin (approved):
Great solution to the two separate app services. ty

PR #2933, opened 2024-12-12
fix: Convert Decimal to int

dhaselhan:
Decimal is not JSON serializable so fails to send message, convert to int before sending


AlexZorkin (approved):
👍

PR #2934, opened 2024-12-16
feat: Sync Reports TFRS -> LCFS

dhaselhan:
* Change logic to sync report with status instead of just transaction
* Call when report is created, submitted, and approved
* Add more fields to message


AlexZorkin (approved):
Looks good thank you. 👍

